PR #14, opened 2014-03-27
Add ability to proxy websocket connections as well as normal HTTP requests

airhorns:
@justinplouffe would you mind giving this a review, and maybe pulling in another stack member?

This adds websocket proxying abilities to the reverse proxy while maintaining all other functionality. We ran into this trying to use the auth proxy in front of ipython notebooks.

 - Adds a `httputil.SingleHostReverseProxy` wrapper which checks the incoming connection for upgrade headers
 - Uses `http` hijacking to take control of the connection if it is for websockets
 - Spawns goroutines to spit data received on the proxy out to the connection to the upstream and vice versa
 - Passes normal http connection to the normal single host reverse proxy
 - Keeps the path based upstream mapping in tact using the `serveMux`

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,9 +5,9 @@ import (
 	"fmt"
 	"log"
 	"net"
-	"os"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 )
 
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -9,7 +9,6 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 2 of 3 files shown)


Manouchehri:
What exactly is still missing from this PR that needs to be done? Quite interested in adding WebSockets support.

discordianfish:
@airhorns Why was this closed?
It seems that bitly doesn't really have interest in maintaining this project properly, given all the dead PRs.
But is there something specifically wrong with this here?

PR #15, opened 2014-04-01
Added support for custom login and error pages

### LINKED ISSUE #11: Allow custom login and error template
The html template is currently hard-coded in `templates.go`. It'd be useful if there was an application switch that allowed specifying a custom template file for both the login and error pages.

hughes:
Fixes #11

I have never written code in golang before, so please review this carefully.

It allows you to pass a `template-path` flag (eg. `--template-path="/var/www/auth"`), which must point to a directory containing `sign_in.html` and `error.html`. If this flag is present, it renders those templates instead of the built-in templates.


frdmn:
+1

jehiah:
custom template support added in #70

Issue #43, opened 2014-11-19
Forward upstream Host header to upstream rather than that of the original request


johnboxall:
First off, thanks for this great piece of software – it fit a need we had very closely and is working great for us with a few modifications.

I'm documenting the changes we made here in case they are helpful for anyone else.

The upstream web services we use google_auth_proxy to protect are available over the public internet at unique hostnames. Each web service is configured to return an `Unauthorized` response for any HTTP request that does not contain a shared secret in a special header.

We've found model to be easily enforceable in our apps and compatible with PaaS which often provide a unique publicly addressable hostname as the only means of addressing a service eg. Heroku.

We wanted to configure a single instance of google_auth_proxy to provide authentication for a number of internal services hosted on Heroku, a PaaS service:

    internal-app-1.company.com => internal-app-1.paas.com
    internal-app-2.company.com => internal-app-2.paas.com
    internal-app-3.company.com => internal-app-3.paas.com

We had difficulty doing this with the current implementation of google_auth_proxy because it:

* Does not set the *expected* Host header on proxied requests. It forwards the Host from the original request rather than using the Host of the upstream.
* Does not provide any mechanism to add additional headers to the upstream request.

Following the above example, when we configured `internal-app-1.company.com` as a google_auth_proxy instance with `internal-app-1.paas.com` as an upstream, a request to `internal-app-1.company.com` made a request to the `internal-app-1.paas.com` service, but with Host header set to `internal-app-1.company.com`.

We changed google_auth_proxy to rewrite the hostnames of requests that it proxies to the upstreams and add the shared secret header to bypass our simple authentication protecting our publicly accessible services.

We've found this to be a useful pattern!

To implement it, we used a custom version of [`httputil.NewSingleHostReverseProxy`](https://github.com/bitly/google_auth_proxy/blob/5199e0028bdde75e919ea950b0c2e9b3fe2a971d/oauthproxy.go#L52) who's [default behaviour is not to rewrite the host header](https://code.google.com/p/go/issues/detail?id=7618) of proxied requests.

If this pattern is something that you find useful, we'd be happy to work to merge it upstream!

PR #45, opened 2014-12-01
#43 – Proxied requests use the Upstream Host Header.

johnboxall:
Hey @jehiah, here is a first take on an implementation of #43.

Could you provide feedback on the following:

1. This change breaks backwards compatibility in original Host header is no longer sent upstream. Would you like me to force the user to pass a flag (as in #1) to enable this new behaviour?
2. For the test, I was inspired by the [`reverseproxy` module's tests](http://golang.org/src/pkg/net/http/httputil/reverseproxy_test.go) to create a Host header echo server. It wasn't immediately clear which hostname I could use that would always to the local server... so I cheated and used xip.io. Any thoughts on how to not involve an this external dependancy?

Thanks!

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -42,7 +42,13 @@ type OauthProxy struct {
 }
 
 func NewReverseProxy(target *url.URL) (proxy *httputil.ReverseProxy) {
-    return httputil.NewSingleHostReverseProxy(target)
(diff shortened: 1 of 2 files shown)


jehiah:
Thanks for the contribution! handled in #68 (with a config option that leaves the existing functionality as default)

PR #51, opened 2014-12-22
Set the default redirect-uri hostname to be the request hostname

mondotron:
Also add a flag to redirect to an https version of that hostname

If your request to google_auth_proxy is the following
    GET /someurl HTTP/1.1
    Host: somedomain.com
and redirect-url is set to ""

redirect-url will be set to http://somedomain.com/oauth2/callback
  or
redirect-url will be set to https://somedomain.com/oauth2/callback
if redirect-to-https flag is set.

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,6 +25,7 @@ func main() {
 
 	flagSet.String("http-address", "127.0.0.1:4180", "<addr>:<port> to listen on for HTTP clients")
 	flagSet.String("redirect-url", "", "the OAuth Redirect URL. ie: \"https://internalapp.yourcompany.com/oauth2/callback\"")
+	flagSet.Bool("redirect-to-https", false, "set default redirect-url to https")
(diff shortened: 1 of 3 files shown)


gavinheavyside:
I vote for this improvement. My use case is simplifying authentication of multiple internal apps behind a single authentication gateway, so I can add in a new app I want to protect without having to run a dedicated google auth proxy for each one, e.g.:

```
                                                ┌────────────┐
         ┌──────────┐      ┌──────────┐    ┌───▶│App1.domain │
         │Google    │      │Load      │    │    ├────────────┤
────────▶│Auth      ├─────▶│Balancer  │────┼───▶│App2.domain │
         │Proxy     │      │          │    │    ├────────────┤
         └──────────┘      └──────────┘    └───▶│App3.domain │
                                                └────────────┘
```

With the change in this PR, I add DNS and load balancer entries, and my new URLs to the project in the Google Console, but I don't need to configure and run and monitor extra Google Auth Proxy processes.

jehiah:
picked up in #69

PR #55, opened 2015-01-07
Add custom SignInMessage as command line parameter

tetsuharu:
This code adds a custom SignInMessage as a command line parameter. 

This allows arbitrary HTML to be passed in via the command line as well. 

The specific use case for this is that I am embedding a google-auth-proxy protected resource in an iframe, and the google auth servers don't allow this type of request to be done inside an iframe. I can work around it by adding some instructions and a link that opens in a new window to the sign in template.


jehiah:
custom template support added in #70

Issue #57, opened 2015-01-19
Option to disable HttpOnly on cookie?


tomtaylor:
I'd like to share a cookie between multiple services, some of which make cross domain requests using JS. Currently, HttpOnly is set to true when the cookie is set and cleared. It'd be great to provide an option to disable this. 

Instinctively, I'd set this using a `-cookie-httponly` flag, but there's already a `-cookie-https-only` flag. I'd be inclined to rename this to `-cookie-secure`, which I think would be closer to the expected naming, but it's a breaking change. 

I'm happy to submit a PR to do this, but I wanted to check your thoughts on renaming the https flag.

tomtaylor:
I just went ahead and submitted the two things as separate PRs.

jehiah:
closed by #58

PR #59, opened 2015-01-19
Rename flag to set Secure flag on cookie

tomtaylor:
As mentioned in #57, I've added this as a separate pull request. I think it makes it clearer, but it's a breaking change.


jehiah:
Can you rebase this now that i merged your other PR? I agree that exposing the `httponly` flag makes the naming for `https-only` awkward.

It probably makes sense to still handle the `cookie-https-only` argument but to just mark it as deprecated

jehiah:
picked up in #71

PR #60, opened 2015-01-19
Run gofmt over source

tomtaylor:
This is just being picky, but gofmt kept complaining about these files, so made a separate PR with the whitespace fixes in.


jehiah:
good catch; when this happens i keep remembering that i need to get something plugged into integration tests that highlights unformatted code.

Issue #28, opened 2014-08-21
group-based authentication


jakeyr:
It would be great to leverage the google apps provisioning API to retrieve the logged-in user's group memberships, enabling some useful features:

  * Restricting access to an application/upstream by group.
  * Passing the list of group memberships foward in a header to the application, allowing for finer grained access control.

vigoatrook:
We are using bitly proxy and it works well for basic OAuth. Now we are trying to get group and orgUnits information and apply cascaded privileges to a user.  The problem with using a file is that the list may not be in sync with Google, since positions can change within the organization, and security concerns, as anyone gains access to the server can change that file. I'm really tempted to implement this feature but like @jakeyr I'm also new to Go. I have a couple of questions from browsing the source. Would you kindly provide some guidance? Thanks

jakeyr:
Since I last looked at this, it appears the google admin SDK has added the ability to enumerate members of a group:

https://developers.google.com/admin-sdk/directory/v1/reference/members/list

it also appears to have been added to the go client:

https://code.google.com/p/google-api-go-client/source/browse/admin/directory_v1/admin-api.json

the new scope is:

https://www.googleapis.com/auth/admin.directory.group.readonly

i can take a crack at implementing this now! hopefully i'll have some time tomorrow or this weekend for this.

jakeyr:
@jehiah can you reopen this issue?

thanks!

jehiah:
This capability was added in #139

PR #61, opened 2015-01-23
Adds User group membership info to headers

jakeyr:
This PR uses the google go admin SDK to determine the group memberships of the authenticating user and stores the list of groups in a signed cookie, and then passes that group list (if valid) forward to the upstream server in the HTTP headers.

See issue #28.

Note that this is my first Go code, so if I'm doing things that are not idiomatic or not organized correctly, please let me know and I can update the PR.

Thanks.


jakeyr:
Note we are currently running this internally and i am running into some issues where certain users seem to not be able to access their own membership lists. it might be better to move to using a "serviceadmin" account instead of per-user Oauth2. I'm investigating.

jakeyr:
ok, i've figured out what's going on.

per this explanation from a google dev, only admin accounts with group api read access can read group membership (even their own, and even after authorizing the "read groups" scope in oauth). so when i tested this with two accounts, they both happened to be admins. everyone else who is not an admin cannot fetch the group list.

so switching this to using a service account is necessary. going to close this PR and reopen when I've got that complete.

thanks.

PR #63, opened 2015-02-11
Added scheme parsing to http-address param

dhowden:
Can now listen for HTTP clients on unix sockets (and any other Go-supported stream oriented network, see http://golang.org/pkg/net/#Listen).  Default behaviour is unchanged, any http-address without a scheme is given the default value of "tcp".

Amended the README so that the usage output is up to date.


jehiah:
great idea!

PR #66, opened 2015-03-16
Catch more options errors at once; add test

mbland:
Notice that the existing `Options.Validate()` has a bug: if the cookie secret is empty, the error doesn't actually get raised.


jehiah:
thanks!

PR #68, opened 2015-03-17
Proxied requests use the upstream Host as Host Header

jehiah:
This extends the work in #45 to add an option for using the request Host Header or upstream host as Host Header for upstream requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 go:
-  - 1.2.2
   - 1.3.3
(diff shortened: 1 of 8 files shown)


johnboxall:
`<3`

jehiah:
@johnboxall If you are able to verify this works for you, that would be awesome

johnboxall:
I'll give it a try this weekend and let you know, thanks @jehiah!

PR #69, opened 2015-03-17
make redirect-uri optional

jehiah:
This constructs a default redirect-url using "http://" + request host header + "/oauth2/callback". This enables easy ability to secure multiple virtual hosts w/ the same GAP instance as outlined in #51


jehiah:
@mondotron @gavinheavyside If either of you are able to validate this change, that would be awesome.

PR #17, opened 2015-03-17
Encoded slashes are expanded by the proxy

jehiah:
To reproduce, hit this URL on the proxy: `http://proxy/a%2Fb/`

It is passed to the backend as `http://backend/a/b/`

The expected result is for the backend to receive `http://backend/a%2Fb/`

This is a problem for Jenkins and is described here: https://wiki.jenkins-ci.org/display/JENKINS/Jenkins+says+my+reverse+proxy+setup+is+broken

### DIFF
--- a/oauthproxy_test.go
+++ b/oauthproxy_test.go
@@ -35,3 +35,30 @@ func TestNewReverseProxy(t *testing.T) {
 		t.Errorf("got body %q; expected %q", g, e)
 	}
 }
+
+func TestEncodedSlashes(t *testing.T) {
+	var seen string
+	backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
+		seen = r.RequestURI
+	}))
+	defer backend.Close()
+
+	b, _ := url.Parse(backend.URL)
+	proxyHandler := NewReverseProxy(b)
+	setProxyDirector(proxyHandler)
+	frontend := httptest.NewServer(proxyHandler)
+	defer frontend.Close()
+
+	f, _ := url.Parse(frontend.URL)
+	encodedPath := "/a%2Fb/"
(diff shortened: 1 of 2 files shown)


jehiah:
@ploxiln mind taking a look at this?

@adrian-gomez I saw you also had a similar patch for this in 3af83e0eef8a545f009d20ade7603ff77fc54e58 so If you have any comments I'd appreciate it.

adrian-gomez:
Hi, yeah i had the issue but with gerrit. The fix as someone mentioned is to use a custom proxy instead of NewSingleHostReverseProx for now at least. The real issue is with NewSingleHostReverseProxy that is not doing a great job as a proxy.

I have fixed this in the commit you mentioned i can do a PR ir you want

jehiah:
@adrian-gomez Yeah your commit was helpful. I've updated this PR w/ a patch and I was hoping you could help verify that a build from this branch works as expected for you.

adrian-gomez:
Yeah np, i'll test and comment back before the end of the day.

oauthproxy_test.go
```diff
@@ -35,3 +35,30 @@ func TestNewReverseProxy(t *testing.T) {
...
+	f, _ := url.Parse(frontend.URL)
+	encodedPath := "/a%2Fb/"
+	getReq := &http.Request{URL: &url.URL{Scheme: "http", Host: f.Host, Opaque: encodedPath}}
+	_, err := http.DefaultClient.Do(getReq)
+	if err != nil {
+		t.Fatalf("err %s", err)
+	}
+	expected := backend.URL + encodedPath
+	if seen != expected {
+		t.Errorf("got bad requst %q expected %q", seen, expected)
```
ploxiln: s/requst/request/

ploxiln:
I'm actually not using google_auth_proxy at the present moment, so take this with a grain of salt, but this LGTM.

PR #73, opened 2015-03-19
Request Logging:

jehiah:
This improves request logging from

```
2015/03/19 20:13:41 127.0.0.1:55370 GET /path/
```

To something closer to Apache Common Log format which includes the Host requested, response code, upstream response timings, etc

```
<REMOTE_ADDRESS> - <user@domain.com> [19/Mar/2015:17:20:19 -0400] <HOST_HEADER> GET <UPSTREAM_HOST> "/path/" HTTP/1.1 "<USER_AGENT>" <RESPONSE_CODE> <RESPONSE_BYTES> <REQUEST_DURATION>
```


ploxiln:
For nginx, you have to configure it to trust `X-Real-IP`, since if it's not behind a load balancer, it might be spoofed by the client. Not a huge problem, just thought I'd mention it...

jehiah:
@ploxiln true, but since GAP doesn't yet support SSL directly i expect it to always be terminated behind something handling SSL ;-)  

Either way i think access logs are much better post this PR

PR #74, opened 2015-03-21
request path contains full URL

jehiah:
This looks like a regression in 1.1.

When I request `/` through the proxy, the upstream receives:

```
GET http://127.0.0.1:8000/ HTTP/1.1
```

The expected request would be:

```
GET / HTTP/1.1
```


jehiah:
@ipmb can you verify that this patch works for you too?

ipmb:
Yep it works. :dancer: 
Thanks!

Issue #65, opened 2015-03-16
Adaptation to other OAuth2 services: Pull requests or forks?


mbland:
I've managed to successfully [fork this repo and adapt it to a new OAuth2 provider](https://github.com/bitly/google_auth_proxy/compare/master...18F:myusa-staging). I know others have done the same. However, given how little work it was, I'm wondering if you'd be open [to generalizing the server to support multiple providers out-of-the-box](https://github.com/bitly/google_auth_proxy/compare/master...18F:generalize), defaulting to Google but allowing for other providers.

I'm happy to take this on, and add tests as I go, as you can see from the latter link. (I'm about to file a PR for a bug I found in `Options.Validate` as part of this process.) That said, I understand if you'd rather keep this instance "pure" and let forks for other providers proliferate as they will.

Apologies if you've address this before; didn't see this come up in any earlier issue. Also, thanks for writing this; we've been using it successfully for months now on our Google-managed domain.

jehiah:
@mbland good questions. Let me dig into your current set of changes to understand what would be involved.

I'm not opposed to some generalization for other oauth provider endpoints and requested scopes, but i suspect it will become tricky as oauth providers don't have a consistent way to echo back identity after authentication which GAP needs.

I'm also keen to eventually support #28 which likely ties some functionality closer to Google's apis.

jehiah:
also cc: @balshor w/r/t 5f8bd41a7c089fdf047380acc9b02f2c19605721

mbland:
OK, I'll keep hacking on [18F/oauth2_proxy](https://github.com/18F/oauth2_proxy) to see if I can come up with some seams that make sense, and we can consider later whether they're worth pushing back up here. Feel free to subscribe to our repo, jump in on PRs, etc.

mbland:
FYI, I've merged the latest changes from here into [the `generalize` branch of our oauth2_proxy fork](https://github.com/bitly/google_auth_proxy/compare/master...18F:generalize?expand=1), and we're running this version now. Seems to work well.

If you're curious, I created a `providers` package to present a unified `providers.Provider` interface between the Google and MyUSA providers, and an `api` package to share the common `api.Request()` (formerly `apiRequest()`) between them.

I still need to add proper tests, but in the off chance you'd be interested in having any of this pushed upstream, let me know and I can prioritize that.

jehiah:
wow. Amazingly that refactoring came out very clean. Totally would accept as a PR. 

[...]

PR #78, opened 2015-03-31
Add ProviderName field; use in sign_in template

mbland:
This is the third step towards genericizing the google_auth_proxy to support OAuth2 providers other than Google as discussed in #65. Injecting the `ProviderName` into the template seems the most expedient way to reuse the default `sign_in.html` template.

Also, @jehiah, I was wondering: If you want, I can follow this PR with a PR to rename the service to `oauth2_proxy`. Otherwise, I'll go ahead and send a PR to add the MyUSA provider (plus README updates on how to add a new provider).

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -279,11 +279,13 @@ func (p *OauthProxy) SignInPage(rw http.ResponseWriter, req *http.Request, code 
 	rw.WriteHeader(code)
 
 	t := struct {
+		ProviderName  string
 		SignInMessage string
 		CustomLogin   bool
 		Redirect      string
 		Version       string
 	}{
+		ProviderName:  p.provider.Data().ProviderName,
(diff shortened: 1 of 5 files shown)


jehiah:
++ let's start with the PR to add MyUSA as a provider. I think i'd like to tackle project & service rename separately and see what other providers make sense to land up front (I know there is an ongoing fork from @balshor to support LinkedIn that I'd love to land using this scheme)

balshor:
I'd love to get LinkedIn support on to this scheme, too!  I'll have another hackday in mid-April that I can use to switch over, if nobody else has added LinkedIn support by then.  :)

jehiah:
awesome @balshor I look forward to that PR!

mbland:
Cool, I'll proceed with the MyUSA+Docs PR, then. Thanks!

PR #80, opened 2015-04-03
Pass the access token to the upstream server

mbland:
@jehiah We're looking to allow other services that accept the `access_token` to be integrated into the 18F Hub per 18F/hub#241. This PR accomplishes this by encoding the `access_token` in the auth cookie and unpacking it as the `X-Forwarded-Access-Token` header for upstream requests. The changes maintain backward compatibility with the original cookie format, but I'm happy to entertain other notions or refactor the code.

@dhcole @afeld @adelevie @yozlet @jackiekazil I've actually got our internal Hub running successfully with these changes to the `google_auth_proxy` and the changes from 18F/hub#241. Just need to merge them to make them official.

I also included a commit to update TestNewReverseProxy(). Without this change, the test will fail when run without an Internet connection. @jehiah I can split that commit out into a separate PR if you prefer.

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -45,6 +47,8 @@ type OauthProxy struct {
 	DisplayHtpasswdForm bool
 	serveMux            http.Handler
 	PassBasicAuth       bool
+	PassAccessToken     bool
+	AesCipher           cipher.Block
 	skipAuthRegex       []string
 	compiledRegex       []*regexp.Regexp
 	templates           *template.Template
@@ -116,6 +120,29 @@ func NewOauthProxy(opts *Options, validator func(string) bool) *OauthProxy {
 
 	log.Printf("Cookie settings: secure (https):%v httponly:%v expiry:%s domain:%s", opts.CookieSecure, opts.CookieHttpOnly, opts.CookieExpire, domain)
 
+	var aes_cipher cipher.Block
+
+	if opts.PassAccessToken == true {
+		valid_cookie_secret_size := false
+		for _, i := range []int{16, 24, 32} {
+			if len(opts.CookieSecret) == i {
+				valid_cookie_secret_size = true
+			}
+		}
+		if valid_cookie_secret_size == false {
+			log.Fatal("cookie_secret must be 16, 24, or 32 bytes " +
+				"to create an AES cipher when " +
+				"pass_access_token == true")
+		}
+
+		var err error
+		aes_cipher, err = aes.NewCipher([]byte(opts.CookieSecret))
+		if err != nil {
+			log.Fatal("error creating AES cipher with "+
+				"pass_access_token == true: %s", err)
+		}
+	}
+
 	return &OauthProxy{
 		CookieKey:      "_oauthproxy",
 		CookieSeed:     opts.CookieSecret,
@@ -136,6 +163,8 @@ func NewOauthProxy(opts *Options, validator func(string) bool) *OauthProxy {
 		skipAuthRegex:      opts.SkipAuthRegex,
 		compiledRegex:      opts.CompiledRegex,
 		PassBasicAuth:      opts.PassBasicAuth,
+		PassAccessToken:    opts.PassAccessToken,
+		AesCipher:          aes_cipher,
 		templates:          loadTemplates(opts.CustomTemplatesDir),
 	}
 }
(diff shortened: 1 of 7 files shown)


oauthproxy_test.go
```diff
@@ -61,3 +65,173 @@ func TestEncodedSlashes(t *testing.T) {
...
+func NewPassAccessTokenTest(opts PassAccessTokenTestOptions) *PassAccessTokenTest {
+	t := &PassAccessTokenTest{}
+
+	t.provider_server = httptest.NewServer(
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			url := r.URL
+			payload := ""
+			switch url.Path {
+			case "/oauth/token":
+				payload = "{\"access_token\": \"my_auth_token\"}"
```
jehiah: Can you backtick this string so it's a little easier to read?

mbland: Done!

oauthproxy.go
```diff
@@ -405,7 +408,11 @@ func (p *OauthProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		// set cookie, or deny
 		if p.Validator(email) {
 			log.Printf("%s authenticating %s completed", remoteAddr, email)
-			p.SetCookie(rw, req, email)
+			if p.PassAccessToken {
+				p.SetCookie(rw, req, email+"|"+access_token)
```
jehiah: Currently cookies are signed (value + timestamp + hmac), but the value is still stored directly in the cookie in base64 encoded format. Previously this was an email address so sensitivity of that information being leaked was relatively low. Storing the access token there changes that. I think we also would need to encrypt the access token with something (the cookie secret?) so that the value itself is also useless.

thoughts?

mbland: Good point. Let me tinker with that.

Two other things that crossed my mind:
- Right now I've got the `PassAccessToken` option set to `true` by default; perhaps I should set it to `false`?
- Should I find a more robust way to concatenate cookie values than just using `+ "|" +`?

jehiah: defaulting to `false` is probably a good choice; I think concatenating with `|` is fine since that's fairly unlikely to occur in an email address or an access token.

mbland:
@jehiah I've just pushed changes to set `pass_access_token` to false by default and to encrypt the access token in the cookie. Confirmed that this is working as advertised on our internal system, in addition to continuing to pass the tests. PTAL?

jehiah:
`PTAL`? that's a new acronym for me

Changes look good. Let's just error if the cookie secret is <32 char and you enabled passing the access token instead of padding with spaces.

mbland:
Ah, PTAL == Please Take Another Look. :-)

Just pushed a new commit with the changes you asked for. (Actually, the AES cipher isn't created if the `cookie_secret` isn't exactly 16, 24, or 32 bytes per [aes.NewCipher](http://golang.org/pkg/crypto/aes/#NewCipher).)

oauthproxy.go
```diff
@@ -405,7 +419,17 @@ func (p *OauthProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		// set cookie, or deny
 		if p.Validator(email) {
 			log.Printf("%s authenticating %s completed", remoteAddr, email)
-			p.SetCookie(rw, req, email)
+			if p.PassAccessToken {
+				access_token, err = encodeAccessToken(p.AesCipher, access_token)
```
jehiah: I think that on the validating request this leaves an encoded access token to be passed in the header because we are re-using this variable.

oauthproxy.go
```diff
@@ -116,6 +120,13 @@ func NewOauthProxy(opts *Options, validator func(string) bool) *OauthProxy {
 
 	log.Printf("Cookie settings: secure (https):%v httponly:%v expiry:%s domain:%s", opts.CookieSecure, opts.CookieHttpOnly, opts.CookieExpire, domain)
 
+	aes_cipher, err := aes.NewCipher([]byte(opts.CookieSecret))
```
jehiah: I'd rather guard this to just when passAccessToken is enabled, and fail hard (`log.Fatal` is probably easiest) if we can't create the cipher instead of falling back to disabled. I think this log should hint at the extra requirements for cookie secret.

jehiah:
Looks good aside from those two small details. Once you've handled those go ahead and squash the commits in this branch down.

oauthproxy.go
mbland: It looks like we always return when we hit this case, but I'll make it a new variable to be on the safe side.

oauthproxy.go
mbland: Sounds good; will do!

oauthproxy.go
jehiah: oh duh. i missed that redirect line in looking at the changes. all good here then.

mbland:
Created a new `encoded_token` variable and cleared `access_token` to be safe; wrapped the cipher creation in a `pass_access_token` case and made an invalid `cookie_secret` size a fatal error; and squashed everything (except the unrelated test fix) down to a single commit.

jehiah:
awesome!

mbland:
_Sweeet_, thanks, @jehiah! Appreciate the thoughtful and timely review.

PR #82, opened 2015-04-07
Redirect to / when /oauth2/sign_in accessed

mbland:
Without this change, clicking the sign-in button on `/oauth2/sign_in` will always redirect back to `/oauth2/sign_in`, essentially creating an infinite loop.

@adelevie @esgoodman @dhcole: I've pushed an instance to the Hub with this feature integrated, and thanks to a comedy of errors, I've already pushed the "Log Out" link to the Hub as well (before issuing a PR as I'd intended). Long and short of it is, there's a working "Log Out" link on the Hub now. :-)


jehiah:
loops are generally bad :angry:

PR #83, opened 2015-04-07
Make address/domain comparisons case-insensitive

mbland:
@jehiah As promised!


mbland:
@jehiah Ping...again, no rush.

PR #87, opened 2015-05-07
Authenticate client using basic authentication.

trecloux:
According to the OAuth2 specification (http://tools.ietf.org/html/rfc6749#section-2.3.1) , the client credentials can be provided using basic authentication or request body.

This pull request adds basic authentication support.

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -205,6 +205,8 @@ func (p *OauthProxy) redeemCode(host, code string) (string, string, error) {
 		return "", "", err
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	b64 := base64.StdEncoding.EncodeToString([]byte(p.clientID + ":" + p.clientSecret))
+	req.Header.Set("Authorization","Basic " + b64)
 	json, err := api.Request(req)
 	if err != nil {
 		log.Printf("failed making request %s", err)


jehiah:
Interesting.

Code structure wise i wonder if this might need to be pushed into the provider code to either add the client id/secret to the post body parameters, or the basic auth header. Since this PR leaves the values in both it's unclear to me what of the 3 supported providers are actually using the Basic auth header.

trecloux:
I think that it is not a problem to send credentials in both ways, the OAuth2 provider can use the post parameters or the basic auth header.

I just tested google provider with basic authentication only and that's ok.

Let me know if you do not want to send both, then I can add a parameter in each provider to drive the client authentication method.

nightlyone:
I think https://github.com/golang/oauth2/blob/master/internal/token.go#L92-L136 may be relevant for this discussion.

jhoblitt:
(Commenting on the feature, not the implementation) This would be a __really nice__ feature to have for interoperability with CLI/scripts.  As an example, the github API will work with basic auth, even when 2FA is enabled, if you generate a personal access token.  At present, it isn't really practical to put this proxy in front of an API end point.

ploxiln:
It's a bit hacky, but that's what the `--htpasswd-file` flag is for - you can manually add some username/password pairs for use with basic auth, for scripts and such. (It was originally used for a nagios mobile app). It does not authenticate with the provider in this case.

PR #89, opened 2015-05-10
Reload authenticated-emails-file upon update

### LINKED ISSUE #41: Allow for dynamic updates of email authorization file
Currently, it looks like the email file is only read at startup. It would be useful to allow this to be read at login so users could be added/deleted dynamically.

mbland:
Closes #41.

The tricky part here is that sometimes popping the file open in an editor and writing to it results not in a direct write, but in some kind of remove and/or rename operation. (At least, that's how Vim appears to roll.) I've confirmed that this works on our running instance, regardless of whether the file is updated in a text editor or, in our case, regenerated via Jekyll (which, oddly, results in a write event, not a remove/rename). Will test out on various Linux flavors and Windows 8.1 a little later today.

cc: @jehiah @razb

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -2,3 +2,4 @@ github.com/BurntSushi/toml              3883ac1ce943878302255f538fce319d23226223
 github.com/bitly/go-simplejson          3378bdcb5cebedcbf8b5750edee28010f128fe24
 github.com/mreiferson/go-options        ee94b57f2fbf116075426f853e5abbcdfeca8b3d
 github.com/bmizerany/assert             e17e99893cb6509f428e1728281c2ad60a6b31e3
+gopkg.in/fsnotify.v1                    v1.2.0
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -40,8 +40,8 @@ func TestRequestFailure(t *testing.T) {
 	resp, err := Request(req)
 	assert.Equal(t, (*simplejson.Json)(nil), resp)
 	assert.NotEqual(t, nil, err)
-	if !strings.HasSuffix(err.Error(), "connection refused") {
-		t.Error("expected error when a connection fails")
+	if !strings.Contains(err.Error(), "refused") {
+		t.Error("expected error when a connection fails: ", err)
 	}
 }
 
--- a/htpasswd.go
+++ b/htpasswd.go
@@ -21,6 +21,7 @@ func NewHtpasswdFromFile(path string) (*HtpasswdFile, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer r.Close()
 	return NewHtpasswd(r)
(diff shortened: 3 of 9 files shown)


mbland:
Also:
- go-fsnotify/fsnotify#17: The exact issue I described above regarding editing the file with Vim. Seems my `time.Sleep()` solution was in line with conventional wisdom, much as I hate to sleep (which is why the `action()` callback argument exists, passed in via `newValidatorImpl()`, so I could use the `updated` channel in `validator_watcher_test.go` to synchronize).
- go-fsnotify/fsnotify#49: I tried watching the directory, i.e. `watcher.Add(filepath.Dir(filename))`,  but my tests aborted with `too many open files`.

Wish I'd read these issues before, but nice to see my findings corroborated. :-)

mbland:
So I found a couple small tweaks needed to be made for Windows 8.1 and pushed a new commit. Also, tests would fail occasionally on Arch Linux; it seems that a `Remove` event is preceded by a `Chmod` on that system. Finally, the tests seemed to deadlock on the Ubuntu 12.04 machines running Travis before the final commit. (The tests take on the order of several seconds to run on Travis now, apparently since there's all this file system activity happening that wasn't before this change.)

After these small tweaks, everything works out on Windows 8.1, Ubuntu 12.04, Ubuntu 14.04, FreeBSD 10.1, FreeBSD 9.3, Debian Squeeze (though I had to change the fsnotify import to use github.com instead of gopkg.in), and Arch Linux.

mbland:
Oh, and OS X 10.10.3, since that's what I normally develop on.

mbland:
Just added `WaitForReplacement()` so watching isn't forever aborted when the authenticated emails file disappears for more than 150ms. It'll call `os.Stat()` up to twenty times a second if it's persistently missing, but that's the pathological case, not the common one.

What's better is that the `os.Stat()` and `watcher.Add()` calls usually succeed right away, and the wait to retry is much smaller if they don't, bringing test times back down to where everybody's happy.

jehiah:
I haven't looked through this closely yet, but at a quick glance this looks great! :star: 

Can you add the new dependency on `gopkg.in/fsnotify.v1` to the `Godeps` file?

mbland:
Done!

BTW, I'm now so intrigued by fsnotify that I forked the repo and am considering coming up with a generic file watcher utility based on `watcher.go` from this PR. We'll see. :-)

jehiah:
:+1: squash down these commits?

mbland:
Done!

mbland:
Hmm, seems there's still a slight bit of a race in the fsnotify.Chmod case. Lemme see...

mbland:
Looks like that did the trick! :-)

PR #92, opened 2015-05-13
Move ValidateToken to Provider, implement LinkedIn correctly

mbland:
Per @balshor's comment in #88, the current ValidateToken() implementation won't work with LinkedIn. This should solve the problem, though I still need to manually test it to be sure.

cc: @jehiah

### DIFF
--- a/providers/linkedin.go
+++ b/providers/linkedin.go
@@ -68,5 +68,5 @@ func (p *LinkedInProvider) GetEmailAddress(unused_auth_response *simplejson.Json
 }
 
 func (p *LinkedInProvider) ValidateToken(access_token string) bool {
-	return validateToken(p, access_token)
+	return validateToken(p, access_token, nil)
 }
--- a/providers/linkedin.go
+++ b/providers/linkedin.go
@@ -66,3 +66,7 @@ func (p *LinkedInProvider) GetEmailAddress(unused_auth_response *simplejson.Json
 	}
 	return email, nil
 }
+
+func (p *LinkedInProvider) ValidateToken(access_token string) bool {
+	return false
+}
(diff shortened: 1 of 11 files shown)


mbland:
Hmm, seems there's still a little race in the file watcher test. I'll see what I can do about that.

mbland:
Now that #93 is in, I rebased this branch and got Travis to pass. I'll try to grab some LinkedIn app credentials to manually try out these changes--though if @balshor wants to pull this branch and beat me to it, that would be wonderful. ;-)

providers/linkedin.go
```diff
@@ -49,9 +60,9 @@ func (p *LinkedInProvider) GetEmailAddress(unused_auth_response *simplejson.Json
 	if err != nil {
 		return "", err
 	}
-	req.Header.Set("Accept", "application/json")
-	req.Header.Set("x-li-format", "json")
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", access_token))
+	for k, v := range getLinkedInHeaders(access_token) {
```
jehiah: Can you make the return type for a few of these `map[string]string` spots `http.Header`?

mbland: Done. Had to propagate changes down to api.RequestUnparsedResponse(). It seems normally a request is created and headers are added to the request's `Header` field directly; but since a `Header` is a `map[string][]string`, it's pretty easy to `make(http.Header)` and pass it around. Didn't bother making it `*http.Header`, since I think `map` types are pretty cheap to pass around anyway.

jehiah:
thanks @mbland this looks good. Want to squash this down? (you knew i was going to ask that, right?) :wink:

jehiah:
commited as 8471f972e141a44f2319e7c7f35ad10c6195c2fa

mbland:
Ah, see you beat me to it. :-)

PR #93, opened 2015-05-13
Provide graceful shutdown of file watcher in tests

mbland:
This test failure from #92 inspired this change:
https://travis-ci.org/bitly/google_auth_proxy/jobs/62425336

2015/05/13 16:27:33 using authenticated emails file /tmp/test_auth_emails_952353477
2015/05/13 16:27:33 watching /tmp/test_auth_emails_952353477 for updates
2015/05/13 16:27:33 validating: is xyzzy@example.com valid? true
2015/05/13 16:27:33 watching interrupted on event: "/tmp/test_auth_emails_952353477": CHMOD
2015/05/13 16:27:33 watching resumed for /tmp/test_auth_emails_952353477
2015/05/13 16:27:33 reloading after event: "/tmp/test_auth_emails_952353477": CHMOD
2015/05/13 16:27:33 watching interrupted on event: "/tmp/test_auth_emails_952353477": REMOVE
2015/05/13 16:27:33 validating: is xyzzy@example.com valid? false
2015/05/13 16:27:33 watching resumed for /tmp/test_auth_emails_952353477
2015/05/13 16:27:33 reloading after event: "/tmp/test_auth_emails_952353477": REMOVE
2015/05/13 16:27:33 failed opening authenticated-emails-file="/tmp/test_auth_emails_952353477", open /tmp/test_auth_emails_952353477: no such file or directory

I believe that what happened was that the call to reload the file after the
second "reloading after event" lost the race when the test shut down and the
file was removed. This change introduces a `done` channel that ensures
outstanding actions complete and the watcher exits before the test removes the
file.

When this goes in, I'll rebase #92 on it to trigger a new test run.

cc: @jehiah @balshor

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -15,13 +15,13 @@ type UserMap struct {
 	m         unsafe.Pointer
 }
 
-func NewUserMap(usersFile string, onUpdate func()) *UserMap {
+func NewUserMap(usersFile string, done <-chan bool, onUpdate func()) *UserMap {
 	um := &UserMap{usersFile: usersFile}
 	m := make(map[string]bool)
 	atomic.StorePointer(&um.m, unsafe.Pointer(&m))
 	if usersFile != "" {
 		log.Printf("using authenticated emails file %s", usersFile)
-		started := WatchForUpdates(usersFile, func() {
+		started := WatchForUpdates(usersFile, done, func() {
 			um.LoadAuthenticatedEmailsFile()
 			onUpdate()
 		})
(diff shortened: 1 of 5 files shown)


mbland:
OK, now both of these commits combined seemed to do the trick. ;-)

mbland:
Ah, Windows. Turns out it keeps watching a file even after a rename, unlike the Unix-based systems. With this third commit, it seems everybody's happy now.

mbland:
Ping... Kinda hoping to get rid of this flakiness I caused... Reputation to uphold, and all. :-)

PR #98, opened 2015-05-21
Add Github Provider

jehiah:
cc: @intellectronica as you had a similar work inprogress


intellectronica:
Thanks a lot for the heads-up @jehiah !

jehiah:
When restricting auth to an Org or Team, it's currently still restricted by email addreses. That will be resolved after #91 lands

PR #85, opened 2015-05-21
Naming Is Hard - project name change

jehiah:
Thanks @balshor @mbland for prompting changes to make the "Google" in "Google Auth Proxy" less relevant. It's high time to rename this project to something more generic.

I've had a few ideas, but i want to open this up to see what ideas anyone else has.

* `GAP` - I'm not sure who started the acronym version of the name but it's stuck. We could just stick with that and remove it's expanded meaning. (if anyone asks we can just say it stood for GreatAuthProxy all along :wink:)
* `AuthProxy` while descriptive, seems too generic
* `OAuthProxy` surprisingly googleable.
* other ideas?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 google_auth_proxy
+oauth2_proxy
 # Go.gitignore
 # Compiled Object files, Static and Dynamic libs (Shared Objects)
 *.o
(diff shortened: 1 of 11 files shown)


jehiah:
:hammer: it's official

PR #102, opened 2015-05-26
rename example file which is referenced in the README

pidah:
simple README fix


jehiah:
thanks for catching this!

PR #104, opened 2015-05-29
Enabling multiple oauth2_proxy's to exist on the same domain

tonymeng:
This PR enables multiple oauth2_proxy's to operate on the same domain while enforcing different levels of security.

The changes required involve namespacing the web template for the specific oauth2_proxy hosting it and namespacing the cookie's key.

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,7 +44,7 @@ func main() {
 	flagSet.String("htpasswd-file", "", "additionally authenticate against a htpasswd file. Entries must be created with \"htpasswd -s\" for SHA encryption")
 	flagSet.Bool("display-htpasswd-form", true, "display username / password login form if an htpasswd file is provided")
 	flagSet.String("custom-templates-dir", "", "path to custom html templates")
-	flagSet.String("proxy-path", "oauth2", "the url path that this proxy should be nested under")
+	flagSet.String("proxy-prefix", "oauth2", "the url root path that this proxy should be nested under (e.g. /<oauth2>/sign_in)")
 
 	flagSet.String("cookie-secret", "", "the seed string for secure cookies")
 	flagSet.String("cookie-domain", "", "an optional cookie domain to force cookies to (ie: .yourcompany.com)*")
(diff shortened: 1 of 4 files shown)


oauthproxy.go
```diff
@@ -300,7 +302,7 @@ func (p *OauthProxy) SignInPage(rw http.ResponseWriter, req *http.Request, code
 	rw.WriteHeader(code)
 
 	redirect_url := req.URL.RequestURI()
-	if redirect_url == signInPath {
+	if redirect_url == fmt.Sprintf(signInPathFormat, p.ProxyPrefix) {
```
jehiah: Can you move these string formatting blocks to happen once when the `&OauthProxy{...}` is created? You can also drop the const's.

main.go
```diff
@@ -44,6 +44,7 @@ func main() {
 	flagSet.String("htpasswd-file", "", "additionally authenticate against a htpasswd file. Entries must be created with \"htpasswd -s\" for SHA encryption")
 	flagSet.Bool("display-htpasswd-form", true, "display username / password login form if an htpasswd file is provided")
 	flagSet.String("custom-templates-dir", "", "path to custom html templates")
+	flagSet.String("proxy-prefix", "oauth2", "the url root path that this proxy should be nested under (e.g. /<oauth2>/sign_in)")
```
jehiah: let's make the value here `/oauth2` which i think will be a little easier to understand.

jehiah:
@tonymeng some comments on implementation, but this change in general sounds like it adds a bit of flexibility which is nice.

oauthproxy.go
```diff
@@ -131,7 +132,7 @@ func NewOauthProxy(opts *Options, validator func(string) bool) *OauthProxy {
 	}
 
 	return &OauthProxy{
-		CookieKey:      "_oauthproxy",
+		CookieKey:      fmt.Sprintf("_oauthproxy_%s", opts.ProxyPrefix),
```
jehiah: Let's just make `CookieKey` a separate configurable not tied to `ProxyPrefix`

tonymeng:
@jehiah thanks for getting back to me so fast, i've updated it per your comments, back to you :)

jehiah:
@tonymeng Thanks! I've merged this (with a few small docs changes and a fix to the cookie name) as c5ccd43767c99eac5d0ec73913bdcbcc03a27789

PR #107, opened 2015-05-31
Fix validator_test hang on Solaris, Plan 9

mbland:
On these platforms, the `done <- true` statement in during TearDown() was hanging, since the `watcher_unsupported.go` version was never draining the channel. Also took the opportunity to update the WatchForUpdates() signature to not return bool anymore.

I developed this on Solaris 11 running in a VM, and confirmed that it's compatible with OS X. I'll expect it'll work just as well with Linux and Windows, and am trying to figure out how to get Go set up on Plan 9 as we speak.

cc: @jehiah

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -21,13 +21,10 @@ func NewUserMap(usersFile string, done <-chan bool, onUpdate func()) *UserMap {
 	atomic.StorePointer(&um.m, unsafe.Pointer(&m))
 	if usersFile != "" {
 		log.Printf("using authenticated emails file %s", usersFile)
-		started := WatchForUpdates(usersFile, done, func() {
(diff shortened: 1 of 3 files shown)


jehiah:
:+1: anything else you expect to add here before we merge this?

mbland:
Nothing I expect, no. But if I manage to get [git installed on Plan 9](http://blog.gopheracademy.com/advent-2014/wrapping-git/) and I find another problem, I'll be certain to open another PR. ;-)

mbland:
FYI, I finally managed to [build and test on Plan 9](https://mike-bland.com/2015/06/08/getting-go-on-plan-9.html), and all is well. ;-)

PR #115, opened 2015-06-23
Cookie Refresh Improvements

jehiah:
I experienced a few issues with the current cookie refresh handling, that left me puzzled, but primarily i got a an empty cookie when a refresh should have happened.

This refactors that code a bit to fix that, and switches the timeframe for renewal to be since a cookie was set. (which i think is simpler to understand). Since the requirements (less than expiry) didn't change i think this is ok.

cc: @mbland

### DIFF
--- a/cookies.go
+++ b/cookies.go
@@ -15,29 +15,39 @@ import (
 	"time"
 )
 
-func validateCookie(cookie *http.Cookie, seed string) (string, time.Time, bool) {
+func validateCookie(cookie *http.Cookie, seed string, expiration time.Duration) (value string, t time.Time, ok bool) {
 	// value, timestamp, sig
 	parts := strings.Split(cookie.Value, "|")
 	if len(parts) != 3 {
-		return "", time.Unix(0, 0), false
+		return
 	}
 	sig := cookieSignature(seed, cookie.Name, parts[0], parts[1])
 	if checkHmac(parts[2], sig) {
 		ts, err := strconv.Atoi(parts[1])
-		if err == nil && int64(ts) > time.Now().Add(time.Duration(24)*7*time.Hour*-1).Unix() {
+		if err != nil {
+			return
+		}
+		// The expiration timestamp set when the cookie was created
+		// isn't sent back by the browser. Hence, we check whether the
+		// creation timestamp stored in the cookie falls within the
+		// window defined by (Now()-expiration, Now()].
+		t = time.Unix(int64(ts), 0)
+		if t.After(time.Now().Add(expiration*-1)) && t.Before(time.Now().Add(time.Minute*5)) {
 			// it's a valid cookie. now get the contents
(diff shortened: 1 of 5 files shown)


cookies.go
```diff
@@ -15,29 +15,35 @@ import (
...
 	}
 	sig := cookieSignature(seed, cookie.Name, parts[0], parts[1])
 	if checkHmac(parts[2], sig) {
 		ts, err := strconv.Atoi(parts[1])
-		if err == nil && int64(ts) > time.Now().Add(time.Duration(24)*7*time.Hour*-1).Unix() {
+		if err != nil {
+			return
+		}
+		t = time.Unix(int64(ts), 0)
+		if t.After(time.Now().Add(expiration*-1)) && t.Before(time.Now().Add(time.Minute*5)) {
```
mbland: Not sure of the utility of `t.Before(time.Now().Add(time.Minute*5))`. Seems like it will always be true, unless there are clock skew issues between different oauth2_proxy instances running on different machines with the same cookie secret.

oauthproxy.go
```diff
@@ -231,19 +231,17 @@ func (p *OauthProxy) ProcessCookie(rw http.ResponseWriter, req *http.Request) (e
...
 		}
 	}
 	if err != nil {
 		log.Printf(err.Error())
 		ok = false
-	} else if p.CookieRefresh != time.Duration(0) {
-		expires := timestamp.Add(p.CookieExpire)
-		refresh_threshold := time.Now().Add(p.CookieRefresh)
-		if refresh_threshold.Unix() > expires.Unix() {
+	} else if ok && p.CookieRefresh != time.Duration(0) && value != "" {
```
mbland: Per jehiah/oauth2_proxy#1, I've removed the `&& value != ""` condition, as it appears redundant with `ok &&`, especially in light of the validation provided by `TestProcessCookieFailIfRefreshSetAndCookieExpired` from that PR.

oauthproxy_test.go
```diff
@@ -348,13 +348,12 @@ func NewProcessCookieTest(opts ProcessCookieTestOpts) *ProcessCookieTest {
 
 	pc_test.opts = NewOptions()
 	pc_test.opts.Upstreams = append(pc_test.opts.Upstreams, "unused")
-	pc_test.opts.CookieSecret = "foobar"
```
mbland: D'oh! Thanks for cleaning up my sloppiness. :-)

mbland:
Thanks for the heads-up! I agree that this implementation does make more sense than the original. Aside from a couple minor comments, and jehiah/oauth2_proxy#1 that I just sent to merge some regression tests and minor tweaks into this PR, it looks good otherwise.

jehiah:
Thanks for the contributions @mbland and for looking through this.

Trhough some of the logging improvements here I figured out that my real issue is that for Google provider the tokens we are using are "online" so they expire after 60 minutes. #116

PR #117, opened 2015-06-25
Google - continually use refresh token

jehiah:
I was mistaken in my understanding in #116 because i thought that when you used a refresh token, the access token you got back was a long term token. It's not and must be continually refreshed when it expires using the refresh token.


jehiah:
@mbland ready for some feedback on this now. I ended up restructuring a bit more than i expected, but i'm much happier with a bunch of the code now, so that makes me :smiley: 

I split apart the horrendously long `OauthProxy.ServerHTTP()` and it should be straight forward to follow the logic in `OauthProxy.Proxy()` now. 

I'll start some production validation of this change set and update with status of that tomorrow or monday.

mbland:
Looking now... Though it'll take a little while for me to scour through everything, offhand I really like the new cookie and session abstractions and how they're exposed in the provider interface. Much better than my proposed hack to maintain the status quo.

oauthproxy.go
```diff
@@ -230,35 +226,35 @@ func (p *OauthProxy) SetCookie(rw http.ResponseWriter, req *http.Request, val st
...
-func (p *OauthProxy) ProcessCookie(rw http.ResponseWriter, req *http.Request) (email, user, access_token string, ok bool) {
-	var value string
-	var timestamp time.Time
-	cookie, err := req.Cookie(p.CookieName)
-	if err == nil {
-		value, timestamp, ok = validateCookie(cookie, p.CookieSeed, p.CookieExpire)
-		if ok {
-			email, user, access_token, err = parseCookieValue(value, p.AesCipher)
-		}
+func (p *OauthProxy) LoadCookiedSession(req *http.Request) (*providers.SessionState, time.Duration, error) {
```
mbland: Nit: Any reason not to use named return variables as done elsewhere?

oauthproxy.go
```diff
@@ -230,35 +226,35 @@ func (p *OauthProxy) SetCookie(rw http.ResponseWriter, req *http.Request, val st
...
-				p.SetCookie(rw, req, value)
-			}
-		}
+		return nil, age, err
 	}
-	return
+
+	age = time.Now().Truncate(time.Second).Sub(timestamp)
+	return session, age, nil
+
```
mbland: Nit: extra newline. But also a convenient place to say: Early return FTW!

oauthproxy.go
```diff
@@ -230,35 +226,35 @@ func (p *OauthProxy) SetCookie(rw http.ResponseWriter, req *http.Request, val st
...
-			}
-			if ok {
-				p.SetCookie(rw, req, value)
-			}
-		}
+		return nil, age, err
 	}
-	return
+
+	age = time.Now().Truncate(time.Second).Sub(timestamp)
```
mbland: Thank you for writing this line. When I was hacking in the refresh support before, I was kinda following what was there, because I hadn't much experience with the `time` API. But wow, this is so much more readable.

oauthproxy.go
```diff
@@ -344,156 +340,223 @@ func (p *OauthProxy) GetRedirect(req *http.Request) (string, error) {
...
+	s = req.RemoteAddr
+	if req.Header.Get("X-Real-IP") != "" {
+		s += fmt.Sprintf(" (%q)", req.Header.Get("X-Real-IP"))
+	}
+	return
+}
 
-	if req.URL.Path == p.RobotsPath {
+func (p *OauthProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	switch path := req.URL.Path; {
```
mbland: So much better. :-)

oauthproxy.go
```diff
@@ -344,156 +340,223 @@ func (p *OauthProxy) GetRedirect(req *http.Request) (string, error) {
...
-		}
+	user, ok := p.ManualSignIn(rw, req)
+	if ok {
+		session := &providers.SessionState{User: user}
+		p.SaveSession(rw, req, session)
+		http.Redirect(rw, req, redirect, 302)
+	} else {
+		p.SignInPage(rw, req, 200)
+	}
+	return
```
mbland: Not sure: Is `return` strictly necessary for functions that don't have return values? The [language ref](https://golang.org/ref/spec#Return_statements) is ambiguous.

oauthproxy.go
```diff
@@ -344,156 +340,223 @@ func (p *OauthProxy) GetRedirect(req *http.Request) (string, error) {
...
-			redirect = "/"
+func (p *OauthProxy) Proxy(rw http.ResponseWriter, req *http.Request) {
+	var saveSession, clearSession, revalidated bool
+	remoteAddr := getRemoteAddr(req)
+
+	session, sessionAge, err := p.LoadCookiedSession(req)
+	if err != nil {
+		log.Printf("%s %s", remoteAddr, err)
+	}
+	if session != nil && sessionAge > p.CookieRefresh && p.CookieRefresh != time.Duration(0) {
```
mbland: If I'm understanding correctly, if `err == nil`, then `session != nil` should always be true, shouldn't it?

Also, I'd put the `p.CookieRefresh != time.Duration(0)` first, since that's a precondition for `sessionAge > p.CookieRefresh` having any meaning.

providers/google.go
```diff
@@ -122,17 +118,49 @@ func (p *GoogleProvider) Redeem(redirectUrl, code string) (body []byte, token st
...
+		AccessToken:  jsonResponse.AccessToken,
+		ExpiresOn:    time.Now().Add(time.Duration(jsonResponse.ExpiresIn) * time.Second).Truncate(time.Second),
+		RefreshToken: jsonResponse.RefreshToken,
+		Email:        email,
+	}
 	return
 }
 
-func (p *GoogleProvider) redeemRefreshToken(refreshToken string) (token string, err error) {
+func (p *GoogleProvider) RefreshSessionIfNeeded(s *SessionState) (bool, error) {
```
mbland: Named return values?

providers/google.go
```diff
@@ -122,17 +118,49 @@ func (p *GoogleProvider) Redeem(redirectUrl, code string) (body []byte, token st
...
+	if s == nil {
+		return false, nil
+	}
+	if s.ExpiresOn.After(time.Now()) {
+		return false, nil
+	}
+	if s.RefreshToken != "" {
```
mbland: Seems you could collapse these three cases into one conditional and an early return (assuming named return values):

```go
if s == nil || s.ExpiresOn.After(time.Now()) || s.RefreshToken == "" {
        return
}
```

oauthproxy.go
```diff
@@ -344,156 +340,223 @@ func (p *OauthProxy) GetRedirect(req *http.Request) (string, error) {
...
+		log.Printf("%s refreshing %s old session cookie for %s (refresh after %s)", remoteAddr, sessionAge, session, p.CookieRefresh)
+		saveSession = true
+	}
+
+	if ok, err := p.provider.RefreshSessionIfNeeded(session); err != nil {
```
mbland: I'm wondering if `revalidated` could be assigned the first return variable from RefreshSessionIfNeeded().

oauthproxy.go
mbland: Just noticed that the previous semantics of validating that the user is still valid via `p.Validator()` once the refresh window has passed appears to have been dropped. Intentional?

If not, the removed `TestProcessCookie*` tests that check this behavior should be reinstated somehow. Can help with this if need be.

mbland:
OK, that seems enough for one round. :-)

oauthproxy.go
jehiah: It's funny that expressing this in terms of `age` only occurred to me after being into this code for a few days. Agreed it's so much simpler to compare with the allowed age before refresh when expressed this way.

oauthproxy.go
jehiah: was just left over from before i split into individual methods per endpoint. I believe the `return` is needed when you have named return variables though.

oauthproxy.go
jehiah: When a cookie is refreshed, that hits the `saveSession && !revalidated` trigger for `ValidateSessionState()` so that should happen as normal. 

The only subcase is if that request also refreshes the access token, it's considered valid w/o calling `ValidateSessionState()`.

mbland: But `p.Validator()` checks the oauth2_proxy's access list, right? So if you wanted to revoke someone's access once their cookie refreshed (by removing them from the emails file), `p.Validator()` would be the means to do that.

jehiah: oooh i was getting confused between `ValidateSessionState` and `Validator`. I follow now, yeah i agree that's now missing from this section. (also a reminder that naming for these two probably needs a little more work)

jehiah: resolved

PR #119, opened 2015-06-25
Fix #118 by updating the documentation

### LINKED ISSUE #118: OAuth Proxy not accepting external traffic
Hi,

I apologise for this as I'm sure I must be doing something wrong, but here we go.

I downloaded and setup the oauth proxy to use with a Google account. After following the various instructions, I ran a command roughly like:

`./oauth2_proxy -upstream="http://localhost:9000" --client-id="my-client-id.apps.googleusercontent.com" --client-secret="my-secret" --cookie-secret="this is a cookie secret for this server"`.

I then executed `curl localhost:4180/ping` and got a 200 response as expected.

I ensured that I opened up the security groups on my instance and tried to access it from my local machine. But I only got `connection refused`.

Skipping a long story, I ended up compiling the simple Go server which is currently here: https://golang.org/doc/articles/wiki/#tmp_3, with a minor modification to use port 4180 instead of 8080.

And everything worked. I could access my Simple Go Server from both localhost (same machine) and from my own dev machine. Which, I think, rules out any firewalls, security groups etc. But possibly doesn't rule out issues with the Amazon AMI I'm using.

I began digging around in the code for the server, and found that the primary difference is this one:

In `http.go`, the current code executes:

`listener, err := net.Listen(networkType, listenAddr)`, which translated to: `listener, err := net.Listen("tcp", "127.0.0.1:4180")`

Whereas the Simple Go Server executes something equivalent to `net.Listen("tcp", ":4180")`.

Changing the code in `http.go` to:

`listener, err := net.Listen(networkType, ":4180")` fixed the problem.

And equally, changing my Simple Go Server to `net.Listen("tcp", "127.0.0.1:4180")` breaks it. I'm running Go v1.4.2 if that's relevant.

So it's something about how that listener gets setup.

So, I think the bug is either:
1. You can't set -http-address=":4180", which I think would solve my problem.
2. Some documentation about what issue I've run into and how to resolve it.

I'm assuming it's some configuration issue around networking I'm not familiar with and therefore the solution is 2. If someone can point me in the right direction that'd be fantastic!

Thanks!

Ed

edhgoose:
Added documentation about listening on 0.0.0.0 as suggested in #118.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,8 +159,27 @@ The command line to run `oauth2_proxy` in this configuration would look like thi
    --client-secret=...
 ```
 
+2) Configure SSL Termination with Amazon ELB, Google Cloud Platform Load Balancing, etc.
 
-2) Configure SSL Termination with [Nginx](http://nginx.org/) (example config below) or Amazon ELB, or ....
+In this case, SSL termination happens on an external load balancer. Therefore you need to set the `--http-address` option like below:
+
+```bash
+./oauth2_proxy \
+   --email-domain="yourcompany.com"  \
+   --upstream=http://127.0.0.1:8080/ \
+   --tls-cert=/path/to/cert.pem \
+   --tls-key=/path/to/cert.key \
+   --cookie-secret=... \
+   --cookie-secure=true \
+   --provider=... \
+   --client-id=... \
+   --client-secret=...
+   --http-address=0.0.0.0:4180
+```
+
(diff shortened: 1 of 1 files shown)


jehiah:
@edhgoose thanks for jumping in with some doc updates. I've added something similar in 51852c045a74210d6405dcc67fa0af83301e2b0b

PR #120, opened 2015-06-26
Incorporate Google refresh token into auth_token

mbland:
Required a change to Provider.ValidateToken() to return a new token if a refresh token was used.

@jehiah This is what I was thinking in my comments on #116. Still need to test it, and it looks like #117 is perhaps a more complete refactoring and renaming of things, but just wanted to put this on your radar.


mbland:
FWIW, added some tests, and tested successfully with a live running instance.

jehiah:
Thanks for the ping. I'll try to merge some of these tests into #117. 

I'm inclined to land with that branch as i wrap up consolidating the session state which makes it easier to separate cookie refresh from token validation/refresh and to clean up the serialization and deserialization of a session from the cookie. That also gives way to more properly support token expiration separate from cookie expiration. I'll ping you there as i wrap that up in the next day or so.

mbland:
Looks like #117 implements a far more robust solution. :-) Ha! Was just about to comment and close when yours popped up, @jehiah. Happy to take a look whenever you ping.

Issue #125, opened 2015-07-13
Multiple instances, different authorized email addresses


mattk42:
Hey guys,

Had an idea for a unique set up of oauth2-proxy, and it doesn't seem to work. Wanted to get your opinion on it.

I have a couple of different services that I am protecting with this proxy, but I would like to limit them to a different set of email domains/authorized emails depending on the service (so #12 wouldn't do the trick). It would also be nice if a user only had to log in once and would get access to only the tools they are configured for (I know, getting picky).

So, as you might be guessing, I set up multiple instances, with different upstreams and auth configs, but with the same cookie secret. In front of all of these, there is one instance that sites in-front of a little homepage with links to all of the other tools.  The homepage instance allows all for our domain, each of the other instances either allow all from our domain or an authorized_emails list.

The standard case is that someone authenticates with that homepage instance and then clicks on the link to whatever tool they want.

After trying this out, it seems that after authenticating with the first instance it doesn't matter what domain/email my user has, they are allowed through. So I am guessing (I don't know Go very well and haven't spent too much time looking) that the code only checks against the authorized list on initial auth, then as long as you have a valid cookie encrypted with the right secret you are allowed through?

I'm going to do some digging myself, but would it be plausible to have a flag to have the proxy check against the authorized list on every request?

Thanks! This proxy is really great, glad I stumbled on it.

jehiah:
@mattk42 thanks for sharing some thoughts on your use case, and i'm glad you were able to get multiple instances configured and sharing authentication successfully.

There isn't a good reason that the email validation isn't done on each request; it's a pretty cheap operation, so that would be a pretty straight forward change. 

If that was changed, it sounds like things would work as you desire, with the small caveat that the behavior on a request to one of your services with an email that isn't valid would remain as is (ie: that invalid request would kill the session even if it happened to be valid on some of your other instances).

PR #126, opened 2015-07-14
Check email validity on all requests rather than only on login.

mattk42:
I believe this small change handles #125.


jehiah:
thanks!

PR #129, opened 2015-07-24
Add support for setting the basic auth password.

jburnham:
For tools that don't like empty passwords, this change allows
one to set a shared secret password for all users.


jehiah:
Thanks @jburnham. Out of curiosity, what tool were you using that didn't play well with the empty password?

jburnham:
https://www.elastic.co/products/shield

PR #131, opened 2015-07-25
Allow passing the value of "approval_prompt" as a flag or option.

ebardsley:


jehiah:
Hi @ebardsley Thanks for the contribution. 

Can you share what provider you were adding this for? I think the feature could be useful but i need to think about implementation as i imagine different providers handle approval prompts differently.

ebardsley:
I'm using this w/ Google as the provider. Previously, Google Accounts would
prompt me for "offline access" when going through the Oauth flow when I was
already logged into the Google account; with this, it just redirects back
to the protected site and login is "transparent".

jehiah:
@ebardsley Thanks that context is helpful. 

What version of oauth2_proxy were you using? Some recent changes added the ability to transparently refresh sessions (via `--cookie-refresh`) so they are essentially perpetually valid (though enforced that you actually still have a working token). That should remove/eliminate most needs to transparently re-auth which sounds like your goal. (And, as a bonus, cookie-refresh works in cases where a transparent loop through the oauth flow  would not like a POST request).

ebardsley:
I dug into timing a bit, and found that when I first wrote up these changes
(before cleaning them up and sending them out) it was late Apr, before the
cookie-refresh changes went in. The new --cookie-refresh stuff definitely
helps and is useful.

That said, I still occasionally hit the protected site on new devices, or
with browsers that don't keep cookies between sessions, or in incognito
mode; assuming that I'm already logged into a Google account w/ a valid
oauth grant, I'd rather have that part work transparently, thus this added
option.

FYI, the other change, which I was planning to send later, redirects
directly into the flow instead of showing a sign-in page, so between that
change and this, removes all (two) interstitial pages from the login flow
for a site that already has a valid grant for a Google account. (See:
https://github.com/ebardsley/oauth2_proxy/compare/master...ebardsley:redirect
)

jehiah:
That's a good point that even with cookie refresh, this accelerates the auth process on a new device.

fyi: since oauth2_proxy supports custom templates I believe you should be able to accomplish the auto-initiate-oauth flow by using a template that has a meta refresh tag instead of waiting for users to click the sign-in button. You shouldn't need any code changes for that.

ebardsley:
Acknowledged regarding the templates, but that's not part of this pull
request. Is there anything else that you'd like to see to allow customizing
the value of "approval_prompt"?

ebardsley:
@jehiah Is there anything else that you'd like for this pull request?

ebardsley:
Also note that the default behavior ("force") with this change remains unchanged; it simply makes it possible to change approval_prompt if desired.

jehiah:
thanks @ebardsley! sorry for the delay in catching up on this PR

PR #139, opened 2015-08-26
[RDY] google: Support restricting access to a specific group(s)

jburnham:
This change should be backwards compatible and only when you pass the three new flags will Google group validation happen.

partially implements features requested in #28 ( getting a list of all groups a user is a member of does not seems to be a single api call and would require listing all groups and then for all groups, querying the group to see if the user is in that group )

potential future improvements (not planned for this PR)
* support orgunits

### DIFF
--- a/providers/google.go
+++ b/providers/google.go
@@ -6,17 +6,25 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/admin/directory/v1"
 )
 
 type GoogleProvider struct {
 	*ProviderData
 	RedeemRefreshUrl *url.URL
+	// GroupValidator is a function that determines if the passed email is in
+	// the configured Google group.
+	GroupValidator func(string) bool
 }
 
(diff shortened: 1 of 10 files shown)


jehiah:
@jburnham Awesome work. I'll review this in the next day or two and get back to you. I'd also love to know if you are successfully using this in production?

jburnham:
haha, not yet, but there are a few places where it would be useful. Just need to set it up.

jburnham:
I still need to support paging as I found an example in https://github.com/google/google-api-go-client/blob/2c0e3555140502254399772c277c54fbaa207e80/examples/gmail.go#L84-L87

providers/google.go
```diff
@@ -6,17 +6,25 @@ import (
...
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
+
+	"golang.org/x/oauth2"
```
jehiah: @jburnham can you capture these dependencies in the Godeps file?

jburnham: @jehiah Added

jehiah:
@jburnham Thanks!!

@markrechler was able to help validate that these changes worked successfully on the Bitly stack so i think we are ready to land this. Can you squash your changes down to a single commit?

jburnham:
@jehiah Great! I'll squash these down later tonight.

jburnham:
This is ready now. We also found another clear new use-case for this feature that I'm excited to see in place!

Issue #141, opened 2015-09-23
Static file serving


Tenzer:
It would be neat if oauth2_proxy could use a directory path as the backend instead of a HTTP server, ie. for generating documentation to static files which only should be accessible for people who have the right credentials.

It's not an urgent feature for me, but I am thinking it's something that may not involve a lot of work.

jehiah:
@Tenzer interesting. I'd accept a patch that supported `file:///...` as a valid upstream setting. Want to take a pass at that?

(I wrote [`private_s3_httpd`](https://github.com/jehiah/private_s3_httpd) with a similar motivation with the only detail that serving files locally didn't fit my need)

Tenzer:
Sure, I'd like to give it a stab although my experience with Go is somewhat limited. From what I can see by giving the project a quick glance, I need to add some logic in `NewOauthProxy()` in oauthproxy.go to add a `net/http.FileServer` as a handler on the `serverMux` being created. Does that sound correct to you?

jehiah:
@Tenzer exactly. `http.FileServer` would be in place of `UpstreamProxy`  if the upstream protocol is `file:///`

Tenzer:
Cool, I'll give it a shot later today and hopefully come up with a pull request :)

jehiah:
feature added in #142

Issue #12, opened 2014-03-18
Documenation for multiple upstreams?


ekristen:
I've read through the README and looked at the code, although I'm still pretty new to Go. I don't understand how multiple upstreams is suppose to work? Can the upstreams be multiple different domains? Or only unique uri? 

"If multiple, routing is based on path" is a little vague.

Any help would be appreciated.

skippy:
hey folks,
I think I figured out a way to make this work... any thoughts on how to simplify are greatly welcome!  I hope this makes sense :)

1. in https://console.developers.google.com, choose your app, go to `APIs and Auth` => `Credentials` and set your `REDIRECT URIS` and `JAVASCRIPT ORIGINS` to `https://auth.internal.example.com/oauth2/callback`
2. for nginx, I have the following config files (with 2 internal apps; in fact this is protecting many more services, each with their own unique subdomain)
```console
upstream google-auth {
  least_conn;
  server localhost:4180;
}

server {
  listen *:8080;
  server_name ~^auth.(?<domain>internal.*)$;
  location = /oauth2/callback {
    proxy_pass http://google-auth;
  }
  location ~/(?<sub>[^/]+)(?<remaining_uri>.*)$ {
    rewrite ^ $scheme://$sub.$domain$remaining_uri;
  }
}

server {
  listen *:8080;
  server_name ~^(.+).internal.*;
  location = /oauth2/start {
    proxy_pass http://google-auth/oauth2/start?rd=%2F$1;
  }
  location / {
    proxy_pass http://google-auth;
  }
}
```
```console
# send to elasticsearch
upstream es {
  least_conn;
  server localhost:9200
}

server {
  listen *:8081;
  server_name es.internal.*;
  location / {
    proxy_pass  http://es;
    proxy_redirect     off;
  }
}
```
```console
# send to consul
upstream consul {
  least_conn;
  server localhost:8500
}

server {
  listen *:8081;
  server_name consul.internal.*;
  location / {
    proxy_pass  http://consul;
    proxy_redirect     off;
  }
}
```
3. google-auth-proxy commands are:
```console
  --upstream="http://localhost:8081" \
  --redirect-url="https://auth.internal.example.com/oauth2/callback" \
  --google-apps-domain="example.com" \
  --cookie-domain=".internal.example.com" \
  --client-id=$GOOGLE_AUTH_CLIENT_ID \
  --client-secret=$GOOGLE_AUTH_CLIENT_SECRET
```
[...]

PR #142, opened 2015-09-23
Add support for serving static files from a directory

Tenzer:
As discussed in #141, this adds the possibility of hosting static files with oauth2_proxy.

The path should be provided as a file:/// url with the full operating system path, ie. `file:///var/www/static/`.

An alias to where the directory is available as can be specified by appending a fragment (ie. "#/static/") at the end of the URL, otherwise the full file path has to be entered in the URL when requesting the files.

Let me know if you have any comments on this. If not then I'll try to get it added to the README file.

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -82,20 +82,35 @@ func setProxyDirector(proxy *httputil.ReverseProxy) {
 		req.URL.RawQuery = ""
 	}
 }
+func NewFileServer(path string, filesystemPath string) (proxy http.Handler) {
+	return http.StripPrefix(path, http.FileServer(http.Dir(filesystemPath)))
+}
 
 func NewOauthProxy(opts *Options, validator func(string) bool) *OauthProxy {
 	serveMux := http.NewServeMux()
 	for _, u := range opts.proxyUrls {
 		path := u.Path
-		u.Path = ""
-		log.Printf("mapping path %q => upstream %q", path, u)
-		proxy := NewReverseProxy(u)
-		if !opts.PassHostHeader {
-			setProxyUpstreamHostHeader(proxy, u)
-		} else {
-			setProxyDirector(proxy)
+		switch u.Scheme {
+		case "http", "https":
+			u.Path = ""
+			log.Printf("mapping path %q => upstream %q", path, u)
+			proxy := NewReverseProxy(u)
+			if !opts.PassHostHeader {
+				setProxyUpstreamHostHeader(proxy, u)
+			} else {
+				setProxyDirector(proxy)
+			}
+			serveMux.Handle(path, &UpstreamProxy{u.Host, proxy})
+		case "file":
+			if u.Fragment != "" {
+				path = u.Fragment
+			}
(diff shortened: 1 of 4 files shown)


oauthproxy.go
```diff
@@ -82,20 +82,32 @@ func setProxyDirector(proxy *httputil.ReverseProxy) {
...
 	for _, u := range opts.proxyUrls {
 		path := u.Path
-		u.Path = ""
-		log.Printf("mapping path %q => upstream %q", path, u)
-		proxy := NewReverseProxy(u)
-		if !opts.PassHostHeader {
-			setProxyUpstreamHostHeader(proxy, u)
-		} else {
-			setProxyDirector(proxy)
+		if u.Scheme == "http" {
```
jehiah: can you make this a case statement (Go convention is to prefer case over if/else in many cases. pun intended =))

```go
switch u.Scheme {
case "http", "https":
   ...
case "file":
   ...
default:
   panic(fmt.Sprintf("unknown upstream protocol %s", u.Scheme))
}
```

jehiah:
@Tenzer this looks like a great first path. I think using the fragment as the path is clever. 

I need to think a little about what that means for typical configurations, and if there should be a way for us to support relative filesystem paths. In the mean time take a pass at README updates and I can always follow up with a future update for relative paths if needed.

oauthproxy.go
Tenzer: Sure, I have switched to a case statement (I can make puns as well ;)) in a4feea9. The reason why I did the if/else is probably due to I normally work in Python, which doesn't have a switch/case statement.

Tenzer:
Hmm, tests are failing now due to the check of the scheme for the provided upstreams.

I don't think there's an official way of supporting relative paths in file:// URLs. You would instead probably have to make a convention where `file:///./`in the beginning of upstream configuration would be replaced with the current folder. I doubt that would give any problems, since I don't think you can make a folder named `.` in any Unix system.

I'll add some details to the README file.

Tenzer:
I have added a section to the README file now which describes a bit more in general how to configure upstreams as #12 also requested that. There may be changes that has to be made to the text, but we at least have something to base corrections upon now.

Tenzer:
I think the easiest way to fix the failing tests, is simply to remove the lines inside `oauthproxy_test.go` which add an upstream of `"unused"` as `NewOauthProxy()` in `oauthproxy.go` then wont iterate over the upstream and complain about the scheme being off. From what I can see, adding a upstream of `"unused"` doesn't add anything and there's no checking in the tests for if a upstream is provided at all.

jehiah:
@Tenzer that or updating the value to `http://unused/`. Once the tests pass go ahead and squash this down to a single commit and we'll land this PR. The docs look good.

Tenzer:
I opted for removing the lines since they aren't part of what is being tested. In case a later addition would check if the upstream is available before it get added the tests would fail again.

The commits has been squashed and the tests has passed. Thanks for the guidance on adding this!

jehiah:
:star: :rocket: :tada: awesome work!

PR #145, opened 2015-09-27
WebSockets3: the Revenge of WebSockets

grahamrhay:
Rebased #14

### DIFF
--- a/logging_handler.go
+++ b/logging_handler.go
@@ -4,6 +4,8 @@
 package main
 
 import (
+	"bufio"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -24,6 +26,16 @@ type responseLogger struct {
 
 func (l *responseLogger) Header() http.Header {
 	return l.w.Header()
+}
+
+func (l *responseLogger) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	highjacker, ok := l.w.(http.Hijacker)
+
+	if !ok {
+		return nil, nil, errors.New("webserver doesn't support hijacking")
+	}
+
+	return highjacker.Hijack()
 }
 
 func (l *responseLogger) ExtractGAPMetadata() {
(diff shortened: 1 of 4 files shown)


a86c6f7964:
I tried running this code and added ws:// endpoint as an upstream endpoint under the exact same path as my http:// endpoint and got
```
2015/10/15 14:26:36 oauthproxy.go:103: mapping path "/" => upstream "http://ip-172-31-6-32.us-west-2.compute.internal:40000"
2015/10/15 14:26:36 oauthproxy.go:108: mapping web sockets "/" => upstream "ws://ip-172-31-6-32.us-west-2.compute.internal:40000"
panic: http: multiple registrations for /
```
I changed the config to assume that that the websockets is always under a subpath, in this case I am using jupyter notebook so I think that it's under /api/kernels/ .

After applying this patch and changing the config to have different subpaths, it looks like everything is working. I am unsure though if it makes sense to make people specify a specific path like that in the configuration. I think that my original configuration would have been a lot easier for people to use.

Vestaxis:
I couldn't get this to work either for exactly the same reason that you describe. Since I wasn't able to specify different subpaths, and I'm only proxying an app that needs websockets, I changed this in line 105 of oauthproxy.go:

```
serveMux.Handle(path, &WebsocketReverseProxy{u.Host, proxy})
```

In other words, I'm simply always using the WebsocketReverseProxy instead of the UpstreamProxy that was used for http/https traffic. This seems to work fine.

Another suggestion for a less 'brute force' method would be appreciated.

grahamrhay:
hmm, you could probably make that work by using the same proxy for both and sniffing the request scheme

grahamrhay:
here: https://github.com/Gamevy/oauth2_proxy/blob/proxy_websockets/websocket_reverse_proxy.go#L70

a86c6f7964:
I think that is more of what nginx does, they just have an option to allow upgrading

neurogenesis:
+1  looking forward to seeing the new and improved websocket functionality merged.

apsey:
I was having websockets issues with ipython terminal (terminado) when using this solution (#145). After reverting it and adding soellman's implementation (https://github.com/bitly/oauth2_proxy/pull/201), it started working.

PR #147, opened 2015-09-29
Apply HMAC signatures to upstream requests

### LINKED ISSUE #146: Securing authentication across the public internet
Short version: Should we add HMAC signatures when passing authenticated requests upstream?

I've been running [oauth2_proxy to implement single sign-on](https://github.com/18F/hub/blob/master/deploy/SSO.md) across different apps on different domains for some time, with great results. However, all of those different apps and domains are actually running on the same host.

Recently my team tried to use the same mechanism for an app running on a different host, across the public internet, and hit the thorny problem of being able to access the app both through the oauth2_proxy and when bypassing it completely. It just hit me that adding optional HMAC signatures to authenticated requests might do the trick. It would be up to the upstream application to check the signatures, but that's easy to implement.

Thoughts? (I'm already poking around the code.)

mbland:
Closes #146.

This still a work-in-progress, as I need to add tests to verify that the actual signing of requests by the proxy is implemented properly. I will remove the "[WIP]" from the title and ping when that's done.

However, the `signature` package itself and the updating of `options` to include the new `signature_key` and `upstream_keys` configuration values is tested and complete.


mbland:
OK @jehiah, this is ready for a look-see. I haven't squashed down the commits yet, just in case I end up wanting/needing to tweak something. Also, the last commit is a broader change to allow for more algorithms besides sha1; all the commits before that hardcode sha1.

Note that `signature.ValidateRequest` should be immediately usable by Go programs wishing to validate signed requests (I _think_ `go get` will do the right thing). Also, all of `StringToSign`, `RequestSignature`, and `ValidateRequest` should be easy to implement in other languages; I plan to write a Ruby gem and an npm if and when this change makes it in.

mbland:
I've gotten the jump on the Ruby and Node implementations; as expected, they were more or less straight ports of `signature/signature.go`:

https://rubygems.org/gems/oauth2_proxy_authentication
https://www.npmjs.com/package/oauth2-proxy-authentication

I'll get on the Python version soon. Of course, I stand ready to update these and share maintainership based on the outcome of this PR.

mbland:
So I went on a tear, and figured out how generic this HMAC signature logic could be. I've extracted the former `signature` package into [`github.com/18F/hmacauth`](https://github.com/18F/hmacauth), and also created:

https://rubygems.org/gems/hmac_authentication
https://www.npmjs.com/package/hmac-authentication

So now the headers could even be configurable between instances.

jehiah:
@mbland I'm excited for this PR. It certainly adds a few new configuration options for oauth2proxy.

Some general feedback on a first read:

* I'm not convinced there is a compelling need to have separate signatures/secrets for each potential upstream. It feels like it would be much simpler (in implementation and usage) to just have `--signature-"sha1:s3kr3t"` and apply that to all upstream requests. Thoughts?
* The signature in `github.com/a18F/hmacauth` is calculated with `header.Get()` but we need all the headers of a particular header in case they are repeated. I *think* this also implies we need output the header key as part of the signature to disambiguate where values came from.
* The signature has the full URL and i'm pretty sure that means it's requires that you must use `-pass-host-header=true` (the default). I think this is ok, but it's worth documenting as otherwise the upstream wouldn't have enough information to re-construct the signature.
* Header order. Would it be useful to just define that the headers are signed in sorted order?

mbland:
Hey @jehiah:

- I've removed the support for per-upstream secret keys. I've saved a patch in case we ever want to add it back.
- Looking more closely at the documentation for [Header.Get()](https://golang.org/pkg/net/http/#Header.Get), I see now that it says it will return the "first" value. I'll update the 18F/hmacauth package.
- Actually, the intent is to pass just the path + query, so `pass-host-header` shouldn't be necessary. Again, I'll update 18F/hmacauth.
- So you're saying just sort all the headers by name and hash them all? Interesting approach that wouldn't even require a configuration of which headers to pick. Can't think of a downside offhand.

As a government employee, I can't solicit effort, but I'll copy you on the 18F/hmacauth PR(s). :wink:

jehiah:
oops, no i wasn't implying to sort **all** headers, i was implying we sort the order of the headers we are hashing. english--

mbland:
Ah, got it. I'll keep the header config bit, then.

mbland:
FYI, I managed to get a generic proxy server together based on the 18F/hmacauth package. Check it out: https://github.com/18F/hmacproxy/

However, I don't know why the build is breaking; it somehow can't find 18F/hmacauth, even though it's public and specified in the `Godeps` file. Any hints, @jehiah?

mbland:
What d'ya know, it's passing again. :-) @jehiah So long as you're comfortable with the implementation of github.com/18F/hmacauth, this may be ready for another look.

But to pick up two points from before:

- I've implemented the hashing of multiply-defined headers in the hmacauth package, as well as in the [Node.js HMAC auth implementation](https://github.com/18F/hmac-authentication-npm/pull/1), but apparently that's a discouraged practice, and Ruby unilaterally selects the value from the last instance of a header (see the notes in the [README from the Ruby HMAC auth implementation](https://github.com/18F/hmac_authentication_gem/pull/1).

- I've not sorted the headers for hashing, but I'm really on the fence about that. If you think it's really worth it, I can update hmacauth (and the other implementations) accordingly.

mbland:
I just pushed a commit to simulate a request body buffer consumed by Read(), which is why the test is currently failing.

Using fakeNetConn in the test exposes a bug in 18F/hmacauth when handling POST requests, addressed by 18F/hmacauth#4. The bug was that the strings.Reader does not consume its buffer contents the same way that a net.Conn does. So the test would pass because its request body would still be intact after signing, but during live testing, the request body would be consumed by HmacAuth.requestSignature().

It also happened to expose a subsequent 18F/hmacauth bug addressed in 18F/hmacauth#5. It turns out that checking Request.ContentLength is an unreliable way of detecting that a body is present, and checking body != nil is sufficient.

18F/hmacauth#4 is already merged; when 18F/hmacauth#5 is in, I'll tag 18F/hmacauth v1.0.1 and update the Godeps to use that version, at which point the test should pass.

cc: @dlapiduz @jcscottiii

mbland:
OK, hmacauth bumped to v1.0.1 and the test is passing again.

jehiah:
@mbland :smile: I think this is in good shape now! Sorry i've been silent here over the past week. I should be able to help wrap this up today

mbland:
No worries, and no rush! :smile:

mbland:
@jehiah Also rebased this branch after #164.

mbland:
...and after #153. :smile:

jehiah:
merged with some slight readme changes as e7dedeb523ac46d46acf0d4d1bebd745a438f9a4

mbland:
Ah, thanks, @jehiah!

PR #148, opened 2015-10-03
Pass `ProxyPrefix` into the error template.

johnboxall:
The default `error.html` uses `ProxyPrefix` but it isn't supplied in the context, causing it to render incompletely:

image


jehiah:
Thanks!

Issue #149, opened 2015-10-03
Document Shortest Possible 😎 Path for Noobs


johnboxall:
`oauth2_proxy` offers a [number](https://github.com/bitly/oauth2_proxy#command-line-options) of configuration options.

For folks looking to get up and running, quickly, it would be awesome to add a 15 minute "Quickstart" guide.

This right place for this may be a separate repo containing the compiled binaries, and an [`app.json`](https://devcenter.heroku.com/articles/app-json-schema) file to power a "Deploy to Heroku" button.

For others looking to get started quickly, here is how I typically do it using the Google Apps provider and an upstream application which requires a secret HTTP Basic Auth password to access:

0. Create a new repo.

1. Put the latest binaries from the [releases](https://github.com/bitly/oauth2_proxy/releases) pages into your repo.

2. Create an `oauth2_proxy.cfg` file. Here are the minimal options you'll need:
    ```
    # Restrict access to folks with this Google Apps domain.
    email_domains = ["$YOUR_GOOGLE_APPS_DOMAIN"]

    # Send the HTTP Basic Auth password `PASSWORD` to all upstreams for all users.
    basic_auth_password = "PASSWORD"
    pass_basic_auth = true

    # Rather than forwarding the `Host` header of the client to the upstream, use
    # the host defined in the upstream configuration.
    pass_host_header = false

    # The application that you're providing authentication for.
    upstreams = ["https://$YOUR_UPSTREAM_APP.herokuapp.com/"]
    ```

3. Create a Heroku app using the binary builtpack:
    ```sh
    heroku create oauth2_proxy_test
    heroku buildpacks:set https://github.com/ph3nx/heroku-binary-buildpack
    ```

4. Create a `Procfile` to run the binary with your config:    
    ```
    web: bin/oauth2_proxy-2.0.1.linux-amd64.go1.5.1/oauth2_proxy -http-address="0.0.0.0:$PORT" -config='oauth2_proxy.cfg'
    ```

5. Follow the [Google Auth Provider](https://github.com/bitly/oauth2_proxy#google-auth-provider) instructions. Set the required environment config variables on Heroku:
    ```sh
    heroku config:set \
        OAUTH2_PROXY_CLIENT_ID="..." \
        OAUTH2_PROXY_CLIENT_SECRET="..." \
[...]

PR #150, opened 2015-10-04
Example of using with Heroku

johnboxall:
Related to https://github.com/bitly/oauth2_proxy/issues/149 – this change adds a "Deploy to Heroku" button with the minimum required configuration to work out of the box.

Unfortunately supporting the Heroku Go buildpack [requires the vendored Godeps format](https://github.com/heroku/heroku-buildpack-go/blob/530952248ba1270a0fea499300bc0758640eed2d/bin/compile#L75) making this a huge diff.

I'm not saying that this is a great idea, but if someone else is looking to spin this up quickly, here you go :)


jehiah:
@johnboxall Thanks for opening up this PR. I like the idea of supporting `Deploy to Heroku` button but unfortunately i do not intend to merge a change that vendors dependencies by copying them into this repo :cry: :sob: 

As a result, i'm inclined to close this issue. Thoughts?

johnboxall:
@jehiah I think its more here for Google than anything else.

I'll close this issue and link to it from #149.

PR #151, opened 2015-10-05
travis: custom install steps to fix build

jehiah:
This adds a workaround for upstream travis Go image changes which added `install: make get-deps` breaking the way this repo expected dependency management.

noticed in #147 by @mbland

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,11 @@ language: go
 go:
   - 1.3.3
   - 1.4.2
-script:
+  - 1.5.1
+install:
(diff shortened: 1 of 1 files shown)


mbland:
Hmm, this fix is still broken. :-P Still trying to understand what's going on...

jehiah:
yup; the travis build image changed to do vendoring on install and that wasn't working as expected.... might have masked other issues with the godep files.

longer term, i'd like to switch this project to [`gb`](https://github.com/constabulary/gb) but not sure this is the right spot to do that.

PR #153, opened 2015-10-08
Add /auth endpoint to support Nginx's auth_request

### LINKED ISSUE #152: An /auth endpoint to support Nginx's auth_request
I recently discovered the [Nginx `auth_request` directive](http://nginx.org/en/docs/http/ngx_http_auth_request_module.html), which allows a setup like this:

```
server {
  listen 443 ssl spdy;
  server_name  myservice.com;

  include ssl/star.myservice.com.conf;

  location = /auth {
    internal;
    proxy_pass http://127.0.0.1:8080;
    proxy_set_header X-Original-URI $request_uri;
  }

  location / {
    auth_request /auth;
    ...
  }
}
```

Would you be open to having me add an `/auth` endpoint or similar to the oauth2_proxy?

It'd basically involve splitting the authentication logic from `func (p *OauthProxy) Proxy(rw http.ResponseWriter, req *http.Request)` into a new `AuthenticateRequest()` function, adding a thin `AuthenticateOnly()` handler, and registering that with `func (p *OauthProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request)` (and adding the requisite options, of course).

mbland:
Closes #152.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,7 @@ OAuth2 Proxy responds directly to the following endpoints. All other endpoints w
 * /oauth2/sign_in - the login page, which also doubles as a sign out page (it clears cookies)
 * /oauth2/start - a URL that will redirect to start the OAuth cycle
 * /oauth2/callback - the URL used at the end of the OAuth cycle. The oauth app will be configured with this as the callback url.
+* /oauth2/auth - only returns a 202 Accepted response or a 401 Unauthorized response; for use with the [Nginx `auth_request` directive](http://nginx.org/en/docs/http/ngx_http_auth_request_module.html)
 
 ## Logging Format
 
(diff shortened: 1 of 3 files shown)


mbland:
Thinking about it, I realized that `auth_request` can actually pass cookies back through when configured with `auth_request_set`. Just pushed a commit that does more what I first imagined, where the bulk of `Proxy()` has become `Authenticate()`, which is shared between a thin new `Proxy()` and `AuthenticateOnly()`. Tests didn't need to change.

So if a user is logged in, this new commit will allow the cookie to be refreshed and other auth headers to be returned, but the redirect won't happen if the cookie isn't set or if it's expired. In the previous version, _all_ that's returned is 202 or 403.

jehiah:
@mbland I landed some renaming in #164 which conflicts with this. Can you rebase to resolve that, and add some docs for configuring with nginx `auth_request` module.

mbland:
@jehiah Done!

jehiah:
:tada: :cake: :fork_and_knife:

PR #154, opened 2015-10-12
Don’t log date/time on Heroku.

paxan:
Heroku logging infra already took care of time stamping.

With this commit applied, here's how a typical Heroku log stream looks like:

```
2015-10-12T18:57:07.171422+00:00 app[web.1]: oauthproxy.go:96: mapping path "/" => upstream "http://localhost:7777"
2015-10-12T18:57:07.171427+00:00 app[web.1]: oauthproxy.go:122: OauthProxy configured for Google Client ID: 12345.apps.googleusercontent.com
2015-10-12T18:57:07.171434+00:00 app[web.1]: oauthproxy.go:132: Cookie settings: name:__oa2p secure(https):true httponly:true expiry:168h0m0s domain:<default> refresh:after 1h0m0s
2015-10-12T18:57:07.172172+00:00 app[web.1]: http.go:45: HTTP: listening on 0.0.0.0:29025
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,7 +14,11 @@ import (
 )
 
 func main() {
-	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	if os.Getenv("DYNO") != "" {
+		log.SetFlags(log.Lshortfile)
+	} else {
+		log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	}
 	flagSet := flag.NewFlagSet("oauth2_proxy", flag.ExitOnError)
 
 	emailDomains := StringArray{}


jehiah:
@paxan I think this is a fair addition because I do appreciate good Heroku support. Just to be clear though, at some granularity timestamps from an application are more accurate because there can be buffering on the log output before a log management process inserts it's own timestamps.

I think an appropriate config flag for this feature would be something like `-log-include-timestamps=true`.

Interested in updating this?

paxan:
Will do.

On Mon, Nov 9, 2015 at 6:57 AM, Jehiah Czebotar <notifications@github.com>
wrote:

> @paxan <https://github.com/paxan> I think this is a fair addition because
> I do appreciate good Heroku support. Just to be clear though, at some
> granularity timestamps from an application are more accurate because there
> can be buffering on the log output before a log management process inserts
> it's own timestamps.
>
> I think an appropriate config flag for this feature would be something
> like -log-include-timestamps=true.
>
> Interested in updating this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitly/oauth2_proxy/pull/154#issuecomment-155086248>.
>



-- 
Pavel Repin

PR #160, opened 2015-10-28
Redirect Option

cpgravlee:
An option to allow for redirection after authorization to something other than "/".

Potentially breaking change for config files, but needed to be changed to match underscore convention. ProxyPrefix string `flag:"proxy-prefix" cfg:"proxy_prefix"

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ Usage of oauth2_proxy:
   -profile-url="": Profile access endpoint
   -provider="google": OAuth provider
   -proxy-prefix="/oauth2": the url root path that this proxy should be nested under (e.g. /<oauth2>/sign_in)
+  -redirect-option="/portal": "Redirect to somewhere other than root after authorization (e.g. https://internalapp.yourcompany.com/portal)
   -redeem-url="": Token redemption endpoint
   -redirect-url="": the OAuth Redirect URL. ie: "https://internalapp.yourcompany.com/oauth2/callback"
   -request-logging=true: Log requests to stdout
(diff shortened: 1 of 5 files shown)


ploxiln:
You can update the pull request by just pushing additional commits into the branch, you don't need to open a new pull request.

Also, you'll probably be asked to squash trivial/fixup commits together at the end (and the result can be force-pushed to this branch, and will update this pull request, without opening a new one).

jehiah:
@cpgravlee can you help me understand the use case you are targeting? They way oauth2_proxy is written, if you go to `/path` and you are unauthorized, you should end up back at that same URL after authorization.

cpgravlee:
The use case is a nginx instance that references different oauth2_proxy instances for each endpoint.

So say we have /path1 and /path2.  /path1 passes onto oauth2_proxy1 whichi uses /oauth-path1 for its proxy-prefix and needs to direct back to /path1 when authorization is complete.

Without my changes, once authorization is complete it redirects back to /.

rafaelmagu:
We have a different use case where we want to redirect users hitting `/` to `/path1` after login and this would be perfect.

jehiah:
@cpgravlee I don't think this PR is necesary. If you start authentication from `/path1` you should end up there after authentication. `oauth2_proxy` already keeps track of the URL you started authentication from. 

If you have a log trace that shows that isn't working for you or can clarify why that doesn't meet your need that'd be helpful.

twhtanghk:
@jehiah Any hints for the config parameter to redirect user back to the url when user starts authentication from. Once user start authentication form url (e.g. /abc/), oauth2_proxy will end up to redirect to / in my case. Config is attached below.

```
proxy-prefix = "/abc/oauth2"
skip_provider_button = true
http_address = "0.0.0.0:4180"
upstreams = [
  "http://nodeapp.service.consul:1337"
]
provider = "mob"
redirect_url = "https://abc.com/abc/oauth2/callback"
request_logging = true
pass_basic_auth = true
pass_host_header = true
email_domains = [
  "gmail.com"
]
client_id = "id"
client_secret = "secret"
cookie_secret = "0123456789012345"
cookie_secure = false
```

PR #161, opened 2015-11-04
adding option to skip provider button sign_in page

rahdjoudj:
An option to skip the sign_in page and redirect instead to the start page.
That change is for users experience only. My users are fully aware that they will use their google account. They want to avoid any extra steps.


ktheory:
:+1: to this feature. I'd use it too.

Thanks, @rahdjoudj

nhumphreys:
:+1: great feature!

rahdjoudj:
It's busted since last commit. I'll fix that quickly.
Cheers,

rahdjoudj:
Here we go ! back on track :)

PR #163, opened 2015-11-09
providers: add OpenID Connect provider

philips:
See the README for usage with Dex or any other OIDC provider.

This has only been tested against CoreOS Dex but should work against
other providers like Google, Facebook, etc as well.

To try it out see the improved Dex docs here: https://github.com/coreos/dex/pull/176


philips:
Closing in favor of #165

PR #167, opened 2015-11-11
Simplify HTTPS listening, leverage net/http methods

artyom:
Make use of net/http Server.ListenAndServeTLS — this gives upcoming
http2 support in Go 1.6 for free.

### DIFF
--- a/http.go
+++ b/http.go
@@ -59,9 +59,6 @@ func (s *Server) ServeHTTPS() {
 		MinVersion: tls.VersionTLS12,
 		MaxVersion: tls.VersionTLS12,
 	}
-	if config.NextProtos == nil {
-		config.NextProtos = []string{"http/1.1"}
-	}
 
 	var err error
 	config.Certificates = make([]tls.Certificate, 1)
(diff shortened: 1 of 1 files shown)


jehiah:
@artyom can you clarify for me why [`Server.Serve()`](http://tip.golang.org/pkg/net/http/#Server.Serve) wouldn't support http2 automatically?

artyom:
At least current http.Server implementation calls private method [setupHTTP2][] which can only be applied in a meaningful way when tls.Listener is set up by Server.ListenAndServeTLS, since `setupHTTP2` calls [http2ConfigureServer][] performing a lot of low-level setup on Server.TLSConfig which is ignored if Server is started on pre-established TLS connection.

[setupHTTP2]: http://tip.golang.org/src/net/http/server.go?#L2053
[http2ConfigureServer]: http://tip.golang.org/src/net/http/h2_bundle.go?#L2025

jehiah:
Thanks for the pointers. I've not familiarized myself with the hooks/implications w/r/t http2 as that has started to land in golang tip.

I'll try to familiarize myself w/ the implications of http2 support in `net/http` and check back in here soon.

PR #175, opened 2015-11-25
Add Dockerfile that installs versioned dependencies via gpm

colhom:
This dockerfile caches the `gpm install` step unless the `Godeps` file is modified. This makes re-building the container image significantly faster in the common case that dependency information has not changed.


colhom:
ping @philips

PR #177, opened 2015-12-02
Add X-Forwarded-Proto header

### LINKED ISSUE #171: Add X-Forwarded-Proto header when proxying https traffic
Some upstream systems (Jenkins under Jetty for example) need the X-Forwarded-Proto header present to properly function when using oauth2_proxy as both auth and ssl termination. 

Without the header any redirects returned by the upstream are for http instead of https

artyom:
Guess protocol from redirect-url scheme if present, or from cookie-secure flag value.

closes #171

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -601,6 +601,14 @@ func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) int
 	} else {
 		rw.Header().Set("GAP-Auth", session.Email)
 	}
+	switch {
+	case p.redirectURL.Scheme != "":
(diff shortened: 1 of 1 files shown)


mgoldsborough:
@jehiah any idea if/when this would be merged?  This would be helpful to me.  Thanks!

guitmz:
Any news on that? Would be great to have it.

michaelwittig:
I run into the same issue when having Jenkins behind oauth2_proxy. SSL is terminated by nginx before it goes to oauth2_proxy.
Jenkins can not build correct HTTP response header for location if the X-Forwarded-Proto header is not set to indicate that https should be used.

PR #183, opened 2015-12-18
Allow options method to bypass authentication

jburnham:
This allows underlying hosts to set CORS headers as necessary when they receive the OPTIONS request. It seems that the standard when making these preflight CORS requests is to not send authentication data so the oauth cookie is not sent.


jburnham:
Hi, are there are concerns with merging this? We've been running with this patch for quite some time successfully.

Thanks

ploxiln:
re-implemented and merged in #370

PR #184, opened 2015-12-23
Ignore spaces in authorized emails list

stanhu:
Got burned by this one: trailing spaces in the authorized e-mail list would cause a mismatch when a user attempted to log in. This PR trims any spaces from the file.


jehiah:
Good update!

PR #190, opened 2016-01-05
Ldap map

eriste:
This is the code I mention in Issue 185. I'll restate the purpose below.

We're using Google to do oauth2 SSO to an internal service and Ldap to fetch information about group memberships etc. The problem is that some of our users have different Ldap uids and Google email addresses (without domain part), ex eriste != erik.stenvall.

To solve this I propose to have the oauth2_proxy make an Ldap call, searching for a matching email address and then substitute the values X-Forwarded-User and X-Forwarded-Email with the values retrieved from Ldap.

/Ess

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,48 +118,104 @@ An example [oauth2_proxy.cfg](contrib/oauth2_proxy.cfg.example) config file is i
 
 ```
 Usage of oauth2_proxy:
-  -approval-prompt="force": Oauth approval_prompt
-  -authenticated-emails-file="": authenticate against emails via file (one per line)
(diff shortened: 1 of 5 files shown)


jehiah:
@eriste I appreciate your PR, but I don't think this is a common enough situation to land in `oauth2_proxy`, and i'd like to keep this more focused on the oauth flow w/ common providers and proxying to upstreams, rather than a thousand custom options.

I think in general this is better handled by a) doing application side resolution from the google account info against ldap, b) writing your own proxy that stands between oauth2_proxy and your upstream which resolves/validates the account info against ldap, or c) writing your own oauth endpoints (that might themselves kickstart oauth via google) which return the account info after being resolved against ldap.

PR #199, opened 2016-01-25
Add URL fragment to sign-in form

hundt:
This allows a seamless transition for applications that encode
state in the fragment.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -129,6 +129,16 @@ func getTemplates() *template.Template {
 	</form>
 	</div>
 	{{ end }}
+	<script>
+		if (window.location.hash) {
+			(function() {
+				var inputs = document.getElementsByName('rd');
+				for (var i = 0; i < inputs.length; i++) {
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -129,6 +129,16 @@ func getTemplates() *template.Template {
 	</form>
 	</div>
 	{{ end }}
+	<script>
+		if (window.location.hash) {
+			(function() {
+				var inputs = document.getElementsByName('rd');
```
russtacular: This will return any type of element, not just `<input>`, which will cause issues with the use of `.value`. You may want to use `document.querySelectorAll('input[name="rd"]')` if you don't have to worry about ancient browsers.

Originally I was going to comment about iterating through a `NodeList` but I think that's probably not a big deal.

jehiah: ahh. Since this template is pretty simple this should be fine then. thanks for taking a look!

PR #201, opened 2016-01-29
Add websocket support

soellman:
This PR adds support for transparent websocket proxying within the HTTP and HTTPS upstreams.

This method uses gorilla/websocket for both the client and the upstream connections and just bridges the sockets when negotiation of both sides is successful. It also tries to close the connections properly after one side or another is done, although more precise error handling could be done there.

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -1,0 +1,127 @@
+package main
+
+import (
+	"io"
+	"log"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+
+	"github.com/gorilla/websocket"
+)
+
+var (
+	CoookieHeaderKey    = http.CanonicalHeaderKey("cookie")
+	ConnectionHeaderKey = http.CanonicalHeaderKey("connection")
+	SetCookieHeaderKey  = http.CanonicalHeaderKey("set-cookie")
+	UpgradeHeaderKey    = http.CanonicalHeaderKey("upgrade")
+	WSKeyHeaderKey      = http.CanonicalHeaderKey("sec-websocket-key")
+	WSProtocolHeaderKey = http.CanonicalHeaderKey("sec-websocket-protocol")
+	WSVersionHeaderKey  = http.CanonicalHeaderKey("sec-websocket-version")
+
+	ConnectionHeaderValue = "Upgrade"
+	UpgradeHeaderValue    = "websocket"
+
+	HandshakeHeaders = []string{ConnectionHeaderKey, UpgradeHeaderKey, WSVersionHeaderKey, WSKeyHeaderKey}
+	UpgradeHeaders   = []string{SetCookieHeaderKey, WSProtocolHeaderKey}
+)
+
+func (u *UpstreamProxy) handleWebsocket(w http.ResponseWriter, r *http.Request) {
+
+	// Copy request headers and remove websocket handshaking headers
+	// before submitting to the upstream server
+	upstreamHeader := http.Header{}
+	for key, _ := range r.Header {
+		copyHeader(&upstreamHeader, r.Header, key)
+	}
+	for _, header := range HandshakeHeaders {
+		delete(upstreamHeader, header)
+	}
+	upstreamHeader.Set("Host", r.Host)
+
+	// Connect upstream
+	upstreamAddr := u.upstreamWSURL(*r.URL).String()
+	upstream, upstreamResp, err := websocket.DefaultDialer.Dial(upstreamAddr, upstreamHeader)
+	if err != nil {
+		if upstreamResp != nil {
+			log.Printf("dialing upstream websocket failed with code %d: %v", upstreamResp.StatusCode, err)
+		} else {
+			log.Printf("dialing upstream websocket failed: %v", err)
+		}
+		http.Error(w, "websocket unavailable", http.StatusServiceUnavailable)
+		return
+	}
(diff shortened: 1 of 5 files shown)


apsey:
I was having websockets issues with ipython terminal (terminado) when using grahamrhay's solution. After reverting it and adding this implementation, it started working. Thanks!

soellman:
I'm glad it worked for you!

corlettb:
Had a slight problem with this and firefox. Tweaked it to work with:

https://github.com/corlettb/oauth2_proxy/commit/151f03fd4e39f308847e081da526f8a9ba5eb348

soellman:
I've updated the PR with better handling of the upgrade/connection headers - it should handle the firefox case now. Thanks!

jhoblitt:
:+1: I've tested this as working with a websockets app (bokeh) that is broken by v2.0.1.

jhoblitt:
I've dropped a copy of my test build of this PR on s3.

https://s3.amazonaws.com/oauth2_proxy/oauth2_proxy-2.0.1.linux-amd64.go1.5.4.tar.gz
sha1sum is ea8f13932790fb1f1da2a54db9a7b168cf192615

jml:
We're using this branch on our servers and it works great. Would welcome it being merged & released.

soellman:
@jehiah I'd be happy to help in any way get this merged. Please do let me know if there's anything I can do!

jhoblitt:
I am using my binary build above in production.  There have been a few unexpected exits (systemd disables core dumps by default, so it is unknown if it was an abort or clean exit) but after changing systemd to always restart the proxy, the problem has not been observed again.

teh:
Another data point: Tested this with Jupyter notebooks with the Github provider, no SSL, and it works for that case. Code looks OK too.

gurvindersingh:
@jehiah any plans to merge this and add support for websockets ?

Thanks for the nice work :)

nside:
+1

daeyun:
+1

mogthesprog:
Is there any hope of this getting merged into stable anytime soon?  :)

nneul:
FYI: This may be of use to those using this patch in conjunction with apache to handle vhosts. (Would sure be nice to have vhost support in oauth2_proxy too!)

http://stackoverflow.com/questions/22002744/how-to-set-up-an-apache-proxy-for-meteor-sockjs-and-websocket

FrederikNS:
Any progress on this? I could really use Websocket support for a project I'm currently working on.

FrederikNS:
@jehiah Is there any plans to merge this? It looks like multiple people have already tested it.

hgomez-sonarsource:
A very welcomed feature, could it be merged ?

mathewpeterson:
Not that I don't think this should be merged but you can use the [ngix auth_request directive](https://github.com/bitly/oauth2_proxy#configuring-for-use-with-the-nginx-auth_request-directive) in order to make websockets work.

weargoggles:
@jehiah Can you please provide feedback on what needs to be done to make this pull request acceptable?

weargoggles:
In the meantime I've built binaries for the current version, with this patch, and published them here https://github.com/weargoggles/oauth2_proxy/releases/tag/v2.2-websocket

websocket.go
```diff
@@ -0,0 +1,138 @@
...
+	// Copy request headers and remove websocket handshaking headers
+	// before submitting to the upstream server
+	upstreamHeader := http.Header{}
+	for key, _ := range r.Header {
+		copyHeader(&upstreamHeader, r.Header, key)
+	}
+	for _, header := range HandshakeHeaders {
+		delete(upstreamHeader, header)
+	}
+	upstreamHeader.Set("Host", r.Host)
```
groob: I think this also needs `X-Forwarded-Proto`, otherwise a rails app behind the proxy would fail the handshake. Rails has a config.force_ssl setting and it affects ws/wss as well.

websocket.go
```diff
@@ -0,0 +1,138 @@
...
+	SetCookieHeaderKey  = http.CanonicalHeaderKey("set-cookie")
+	UpgradeHeaderKey    = http.CanonicalHeaderKey("upgrade")
+	WSKeyHeaderKey      = http.CanonicalHeaderKey("sec-websocket-key")
+	WSProtocolHeaderKey = http.CanonicalHeaderKey("sec-websocket-protocol")
+	WSVersionHeaderKey  = http.CanonicalHeaderKey("sec-websocket-version")
+
+	ConnectionHeaderValue = "Upgrade"
+	UpgradeHeaderValue    = "websocket"
+
+	HandshakeHeaders = []string{ConnectionHeaderKey, UpgradeHeaderKey, WSVersionHeaderKey, WSKeyHeaderKey}
```
jbandlow: I needed to add "sec-websocket-extensions" to this list to get it to work. (See https://github.com/gorilla/websocket/blob/a91eba7f97777409bc2c443f5534d41dd20c5720/client.go#L237)

paulreimer: Nice find, I also needed this!

mizzao:
Are there any options that need to be turned on for this to work? I'm getting the following error when trying to use WS, though normal HTTP works:

```
'wss://example.com/sockjs/535/hz93j02u/websocket' failed: One or more reserved bits are on: reserved1 = 1, reserved2 = 0, reserved3 = 0
```

PR #206, opened 2016-02-09
add basic-auth-username

hhsnopek:
Adds the ability to set a static username, overriding `session.User`


raphink:
This would be quite useful.

raphink:
What's the status of this @jehiah ?

msh100:
Can we get an update on this? This would be quite a useful addition

mritzmann:
+1 This would be a nice addition.

abl:
Sorry to revive an old PR but I was about to implement this myself! @jehiah, would it be helpful if I took this and fixed the conflicts or do you feel it isn't a feature worth merging?

(My motivation: I don't want to add users in two places - once in the proxy and then a matching record in the backend service.)

PR #208, opened 2016-02-12
Added SystemD example service file

sgnn7:
Since most people want to run this as a service, having some sort of an
example file is very useful to have.


jehiah:
thanks!

PR #211, opened 2016-02-16
Add GitLab provider

### LINKED ISSUE #200: support for gitlab provider ?
Adding support for gitlab as a provider would be great !
http://doc.gitlab.com/ce/integration/oauth_provider.html
thus we could use it with our self-hosted gitlab service.

pmosbach:
Works with either Gitlab.com or self-hosted GitLab instances.

Closes #200.


jehiah:
@pmosbach this looks great can you squash your commits down to a single commit?

pmosbach:
@jehiah certainly

PR #222, opened 2016-02-29
Add more env variable :

samuel-soubeyran:
- allow more configuration from env variable (ProxyPrefix, HttpAddress, HttpsAddress, RedirectUrl, TLSCertFile, TLSKeyFile, )
  - add EmailDomain to env variable
  - add Upstream parameter to env variable


samuel-soubeyran:
I need to configure via env variable a bunch of stuff.
EmailDomain and Upstream break the existing behavior but I think it is also more coherent (there was a singular/plural mismatch between configuration file/ flag name).

jehiah:
@samuel-soubeyran in general this looks like a good idea.

Did you run into a particular problem with specifying a comma separated set of values for `OAUTH2_PROXY_UPSTREAM`? without adding a separate config flag for that? I would expect that to be handled transparently by the code the package [`go-options`](https://github.com/mreiferson/go-options/blob/master/options.go#L160-L164) which is used to combine environment variables w/ the config struct.

samuel-soubeyran:
Good question, I didn't try because I just needed one upstream. And I couldn't make it works with only one. I'll try with multiple and see how it works

samuel-soubeyran:
OK I did some test, It does work for several upstream url. For single upstream url I tried:
"OAUTH2_PROXY_UPSTREAMS=https://xxxxxx.net," (notice the coma at the end).
It crashes:
2016/02/29 20:27:14 oauthproxy.go:123: mapping path "/" => upstream "https://xxxxxx.net"
panic: unknown upstream protocol
goroutine 1 [running]:
panic(0x8141c0, 0xc82010c5b0)
	/usr/local/go/src/runtime/panic.go:464 +0x3e6
main.NewOAuthProxy(0xc8200bc840, 0xc820059620, 0x2)
	/go/src/github.com/revinate-docker/oauth2_proxy/oauthproxy.go:140 +0x25e8
main.main()

I don't know if you have any other idea to make it works for single upstream?

PR #226, opened 2016-03-14
Expand auth_request documentation

rcoup:
Fix missing `/oauth2` prefix on URLs with default proxy-prefix
Expand `error_page` clause to a working example
Add version note (isn't in latest release)
Add `oauth2_proxy` example command config


fuhrysteve:
:+1: this section needs better documentation

PR #207, opened 2016-02-09
Add Cloudfoundry provider

omazhary:
this adds a provider for Cloudfoundry's UAA [1]

if you think this makes sense we can also add some unit tests.

The second commit allows to set X-Forwarded-User independent of HTTP Basic Auth as we need this for our Cloudfoundry/grafana scenario [2]

[1] https://github.com/cloudfoundry/uaa
[2] http://docs.grafana.org/installation/configuration/#authproxy

PR #234, opened 2016-04-11
Added support for self-signed SSL certificates

omazhary:


omazhary:
Branch rebased and ready for merging.

jehiah:
@omazhary can you describe your motivations for Self-signed cert capability?

omazhary:
It's only for development environments where no one uses CA signed certs. However, once it goes productive, SSL (strict) gets re-enabled.

Regards,
Omar Elazhary

Jehiah Czebotar <notifications@github.com> wrote:

@omazhary can you describe your motivations for Self-signed cert capability?

jehiah:
@omazhary so you are trying to set `InsecureSkipVerify` just for the upstream proxied requests? Asking because as written this would also make all communication w/ an authentication provider insecure.

omazhary:
@jehiah nope, all requests should be insecure. If the upstream resource has a self-signed cert, it doesn't work. The same happens with the authentication provider.

For instance, we're testing this with grafana as an upstream resource, and using cloudfoundry's UAA as an authentication provider (hence PR #207). In our dev environments, https is enabled and we use self-signed certs. It makes no sense to invest in CA signed certs for development purposes. However, once the end result gets shipped to a productive environment, strict SSL (overall) is used. Meaning, both the upstream resource as well as the authentication provider are using CA signed certs.

That's why I thought it would make sense to have such an option, to be able to switch it on and off as needed based on your use case.

jehiah:
Thanks. Helpful context. One last question, Did you evaluate l valid free certs through lets encrypt or startcom? I understand the desire not to pay for development certificates but I'm on the fence here because I am concerned about insecure production configurations. 

https://letsencrypt.org
https://www.startssl.com

omazhary:
We did, but since our dev environment is internal, and can't be reached externally, it didn't seem necessary.

PR #241, opened 2016-04-21
Allow setting github provider /user/orgs endpoint url

hartzler:
This is a bit of a hack, but it does allow enterprise github setups to limit to orgs.  Open to recommendation how to do this nicer...


gkuchta:
:+1:

rremer:
@hartzler you may want to fetch upstream and repush, I suspect #249 will fix your ci build.

I think this pr is a good start, e.g. most of the uris could use configurable overrides with sane defaults. I ran into a similar issue with just the base github enterprise api uri in #248 , which is similar (bu not conflicting).

jehiah:
fixed in #269

Issue #144, opened 2015-09-26
Custom github url not supported


meatheadmike:
I've set custom validate, redeem and login URL's so that I can auth against our enterprise github installation. But the oauth proxy appears to ignore these settings and instead tries to validate against public github.

PR #248, opened 2016-05-12
allow github enterprise endpoints with api uris not on root url

rremer:
#144 Fixed the problem where you couldn't specific 'api.mydomain.com' for a github enterprise (ghe) installation, but the default ghe api endpoint is normally mydomain.com/api/$version (e.g. 'v3).

In the existing implementation, oauth2_proxy will incorrectly validate against the configured validation url *without* the configured uri, so:
mydomain.com/api/v3 becomes mydomain.com/user/emails

This pull requests honors the full configured url as a base path, so:
mydomain.com/api/v3 becomes mydomain.com/api/v3/user/emails

Additionally, I reduced the logging on success for validation so we don't printy the entire response body and just log that we got a 200 (like we do for all other requests where we were able to successful unmarshal responses).

### DIFF
--- a/providers/github.go
+++ b/providers/github.go
@@ -64,7 +64,7 @@ func (p *GitHubProvider) hasOrg(accessToken string) (bool, error) {
 		"limit":        {"100"},
 	}
 
-	endpoint := p.ValidateURL.Scheme + "://"  + p.ValidateURL.Host + "/user/orgs?" + params.Encode()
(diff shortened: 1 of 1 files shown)


rremer:
@jehiah Thanks for fixing the travis build, I almost forgot that this was the original bug I had intended to fix (Just updated from upstream).

jehiah:
fixed in #269

PR #251, opened 2016-05-15
Remove superfluous quotes from config path

bluecmd:
Using -config="path" will make the application try to open paths with the quotes included which is not what is expected.

```
main.go:89: ERROR: failed to load config file "/etc/oauth2_proxy.cfg" - open "/etc/oauth2_proxy.cfg": no such file or directory
```


rremer:
@bluecmd you may want to fetch upstream and repush, I suspect your build will be fixed by #249

PR #256, opened 2016-05-25
Add option to set custom footer text

### LINKED ISSUE #166: option to hide version number
Currently oath2_proxy shows the version number on the login page

It would be great to be able to hide this

dustywilson:
Intended to resolve #166.  Adds an option (in all the usual spots) called "footer" that allows for overriding the existing footer text.  It's a static value.  If you want the footer to go away altogether, just set it to " " (single space, not empty string).  If it's set to empty string (default), it'll do the usual version/link footer.  I expect that it's simple plaintext and any HTML would be escaped by the template.


dustywilson:
I don't believe the above CI test failures are due to my patch.  Other PRs near the same time were giving the same error, and tests run fine on my machines.  Please complain to me if I'm wrong about that.  I'm using this patch in production with great success.

rremer:
@dustywilson fetch upstream and push again, I believe you (like me) were building with a recent version of golang which travis was not. Some upstream deps were no longer available for those ancient versions of golang. The travis build works now (see #249 ), so I suspect your build will pass once you rebase.

PR #257, opened 2016-05-26
fix dist.sh compatibility with GNU coreutils mktemp

jhoblitt:
Resolves this error:

    mktemp: too few X's in template ‘oauth2_proxy’

### DIFF
--- a/dist.sh
+++ b/dist.sh
@@ -20,7 +20,7 @@ echo "... running tests"
 
 for os in windows linux darwin; do
     echo "... building v$version for $os/$arch"
-    BUILD=$(mktemp -d -t oauth2_proxy)
+    BUILD=$(mktemp -d ${TMPDIR:-/tmp}/oauth2_proxy.XXXXXX)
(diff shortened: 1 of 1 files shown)


ploxiln:
I suggest a slight variation which works more similarly between OS X and linux:

```
mktemp -d ${TMPDIR:-/tmp}/oauth2_proxy.XXXXXX
```

jhoblitt:
@ploxiln That sounds reasonable.  I went digging through $day_job code and last time I ran into coreutil/osx `mktemp` issues, I ended up with. `$(mktemp -t "${conda_packages}.XXXXXXXX")`  I don't recall why there are 8 `X`s.  Do you know if el cap is also happy with 6 in the pattern?

ploxiln:
I tested on OS X 10.10 (yosemite) and ubuntu 14.04. Looks like coreutils demands at least 3 and OS X doesn't complain for any number of them.

jhoblitt:
I have confirmation from coworkers with el cap that 6 'X"s succeeds.  I'll update the PR in a moment.

PR #227, opened 2016-06-20
Cookie secret encoding?

jehiah:
What sort of encoding is expected for the cookie secret value?  It looks like the validate test is counting the number of characters (16, 24, or 32) in the string but I'm not sure how that relates to the number of bytes represented by the string.

If the encoding was urlsafe base64 a 32 byte string would look like this:

```
GOW9RCRecupnZ7cgGy9wMEGZ3Ssm2SpyAFUx_olbU90=
```

This string is 44 characters with the padding.  Each character represents 6 bits so we need all the bits from the first 42 characters and then 4 bits from the 43rd character, and then the rest is padding.

How are we getting 32 characters == 32 bytes?

### DIFF
--- a/options.go
+++ b/options.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"crypto"
+	"encoding/base64"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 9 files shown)


options.go
```diff
@@ -251,3 +260,26 @@ func parseSignatureKey(o *Options, msgs []string) []string {
...
+	default:
+		return secret
+	}
+}
+
+// secretBytes attempts to base64 decode the secret, if that fails it treats the secret as binary
+func secretBytes(secret string) []byte {
+	b, err := base64.URLEncoding.DecodeString(addPadding(secret))
+	if err == nil {
+		return []byte(addPadding(string(b)))
```
ploxiln: why `=` pad the binary result of base64 decoding?

options.go
```diff
@@ -251,3 +260,26 @@ func parseSignatureKey(o *Options, msgs []string) []string {
 	}
 	return msgs
 }
+
+func addPadding(secret string) string {
+	padding := len(secret) % 4
+	switch padding {
+	case 1:
+		return secret + "==="
```
ploxiln: This case will never happen for a valid base64 encoded string.

PR #197, opened 2016-01-21
use Github provider with GitHub enterprise

ruta-goomba:
Pull request to resolve issue "Custom github url not supported #144"

I removed hard-coded github.com endpoints. With the proposed changes oauth2_proxy works with enterprise github if user sets validateURL (as well as loginURL and redeemURL) to github enterprise endpoint.

PR #269, opened 2016-06-20
github enterprise provider

### LINKED ISSUE #144: Custom github url not supported

jehiah:
This fixes github enterprise support originally added in #197 and closes #144 and #248


rremer:
@jehiah I can confirm that this implementation resolves #248 , thanks!

jehiah:
@rremer 🎉  🎺  🎈  yay! thanks for the confirmation 💯 !

PR #193, opened 2016-06-23
A new tag

jehiah:
Hi guys,

I've noticed theres been quite a few commits since the last tag, and I'd especially like to use the `file://` upstreams without having to rely on an active branch. So with that in mind, do you think a new release could be tagged sometime soon?

Thanks!


tanuck:
👍 👍 👍 

Awesome, thanks a lot @jehiah

0815fox:
Thanks, man, I owe you a beer!

PR #271, opened 2016-06-23
avoid accidentally-base64 cookie secrets being treated as base64

ploxiln:
don't pad result of base64 decoding, this makes some secret strings valid if considered base64, when they shouldn't be

check length of base64 decoding, don't treat as base64 if not valid secret length

No base64 string will have/need '===' padding.

Fix two tests that were interpreted as base64, but only described 15 bytes if so interpreted
- make one 16 chars (not base64)
- make one base64 of 16 bytes (needs 2 more non-pad chars) (and optional pad)


ploxiln:
This could fix a (minor?) regression, where someone's old cookie-secret unintentionally was now considered base64 and thus changed, making old cookies invalid, causing extra re-auth.
But because there was a release, you could consider it another kinda-backwards-incompatible change :)

PR #275, opened 2016-07-01
add support for basic auth user instead of oauthed user with basic auth

rremer:
I was surprised to find out that when authenticating to an upstream there were only the following supported configurations:
a) pass X-Forwarded-User and X-Forwarded-Email
b) pass X-Forwarded-User and X-Forwarded-Email with a configured basic auth password (-basic-auth-password)
c) pass nothing

A common usecase I've run into is passing full basic auth to an upstream (with credentials that are machine-generated and no one sees), and presenting oauth to an end-user. In that scenario, the above options aren't particularly useful. This PR adds a parameter to allow passing a basic auth password AND user to an upstream. We can still pass X-Forwarded-User and X-Forwarded-Email as additional headers, but the username passed with the auth header is not just a duplicate of the X-Forwarded-User.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ Usage of oauth2_proxy:
   -approval-prompt="force": Oauth approval_prompt
   -authenticated-emails-file="": authenticate against emails via file (one per line)
   -azure-tenant="common": go to a tenant-specific or common (tenant-independent) endpoint.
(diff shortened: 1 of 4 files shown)


rremer:
Didn't see #206 for some reason, closing this out since his has a nice test with it.

PR #278, opened 2016-07-17
Validate cookie name 

### LINKED ISSUE #264: No errors logged when Set-cookie header is not sent in response due to an invalid cookie
oauth2_proxy currently will silently fail to complete authentication if an invalid cookie name is set, perpetually redirecting the user back to the login screen with no errors logged.

This seems to be due to the behavior of Go's http.SetCookie method which according to the docs: ` The provided cookie must have a valid Name. Invalid cookies may be silently dropped.`. Since oauth2_proxy's authentication relies on setting a cookie, there should be validation that the `MakeCookie` method makes a valid cookie, and subsequently, that the Set-Cookie header is actually included in the response headers.

tanuck:
This fixes #264 

`(http.Cookie).String()` will silently return an empty string when the cookie name contains a rune not in the following table https://github.com/golang/net/blob/master/lex/httplex/httplex.go#L17

With this patch I've added an additional validator to check the cookie name at boot. I've had to 'mock' a `http.Cookie` as the http package doesn't export the check function, here: https://github.com/golang/go/blob/master/src/net/http/cookie.go#L368

Regards,
James

### DIFF
--- a/options.go
+++ b/options.go
@@ -4,6 +4,7 @@ import (
 	"crypto"
 	"encoding/base64"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
 	"regexp"
@@ -200,6 +201,7 @@ func (o *Options) Validate() error {
 	}
 
 	msgs = parseSignatureKey(o, msgs)
+	msgs = validateCookieName(o, msgs)
 
 	if len(msgs) != 0 {
 		return fmt.Errorf("Invalid configuration:\n  %s",
@@ -257,6 +259,14 @@ func parseSignatureKey(o *Options, msgs []string) []string {
 			o.SignatureKey)
 	} else {
 		o.signatureData = &SignatureData{hash, secretKey}
+	}
+	return msgs
+}
+
+func validateCookieName(o *Options, msgs []string) []string {
+	cookie := &http.Cookie{Name: o.CookieName}
+	if cookie.String() == "" {
+		return append(msgs, "invalid cookie name: "+o.CookieName)
 	}
 	return msgs
 }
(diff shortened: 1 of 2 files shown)


ploxiln:
Not a bad idea.

Unrelated quick tip: your links are broken. It's because you used the "Add a link" button in the github comment editor, and pasted the url into the title part (between `[]`) and left the url part blank (between `()`). Anyway, you can just paste the link in without anything special and it will do as you expect.

tanuck:
@ploxiln thanks for spotting that. Should have double checked...

options.go
```diff
@@ -261,6 +263,14 @@ func parseSignatureKey(o *Options, msgs []string) []string {
 	return msgs
 }
 
+func validateCookieName(o *Options, msgs []string) []string {
+	cookie := &http.Cookie{Name: o.CookieName}
+	if cookie.String() == "" {
+		return append(msgs, "invalid cookie name: "+o.CookieName)
```
jehiah: Can you use `fmt.Sprintf` verb `%q` here?

jehiah:
@tanuck thanks for your contribution. I just had one small comment before we merge this

options.go
tanuck: Yeah sure... Should the usage me made a bit more consistent? I.e `func parseSignatureKey` uses concatenation, whereas others use `fmt.Sprintf`?

jehiah: Let's leave this PR scoped to this message.  

concatenation is "OK" (it's not going to panic or anything) it's just that when the purpose of an error is to highlight unusual values (like a null byte, or a `\r`) `%q` can greatly clarify things vs `%s`.

PR #280, opened 2016-07-21
Add Dockerfile

odedlaz:
The Dockerfile fetches the latest GitHub release and builds it.
This might be better than building straight from source, and makes the build process IMO.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,39 @@
+FROM golang:1.6-alpine
+
+# install su-exec (a lightweight gosu)
+RUN apk add --no-cache su-exec libcap
+
+# install all build dependencies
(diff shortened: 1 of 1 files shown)


underyx:
👎, if I `git checkout` an old version and run `docker build .`, I want (and would expect) that old version to be built.

ahmetb:
@odedlaz would you consider editing this to just build from the source code? It would be easier to take in.

dstroot:
Take a look at this alternative: https://github.com/bitly/oauth2_proxy/pull/372

pataquets:
+1 for [Automated Builds](https://docs.docker.com/docker-hub/builds/) straight from source on Docker Hub.

PR #290, opened 2016-07-31
Adding skip-provider-button docs

MarkHerhold:


jehiah:
Thanks! this was missed in #161

PR #292, opened 2016-08-03
Strip sensitive URL parameters from provider log output

### LINKED ISSUE #291: oauth2 access token is printed to log by default
After setting a Cookie refresh value and setting `--pass-access-token` to true, we've noticed some logs come through that contain the access token, with no apparent way to turn it off.

See: https://github.com/bitly/oauth2_proxy/blob/master/providers/internal_util.go#L31

The `endpoint` var gets set to the fully qualified URL, which contains the `access_token` query parameter. For us, this doesn't pose a security threat as the default (GitHub) token scopes are `user:email` and `read:org`, both read-only.

But, if our application were to request a much larger scope, such as `repo` (full read/write to repositories) this would become a larger issue, especially once private repos get involved.

Am I missing some configuration here? It seems like an option should exist to turn this output off.

nickmiller-wf:
Resolves #291 

After searching through the providers package manually, I determined the GitHub provider and `validateToken` helper function were the only areas at risk here. Other providers, like Google, encode their sensitive parameters as a form-encoded body, and they are never logged.

@jehiah

### DIFF
--- a/providers/github.go
+++ b/providers/github.go
@@ -85,7 +85,8 @@ func (p *GitHubProvider) hasOrg(accessToken string) (bool, error) {
 		return false, err
 	}
 	if resp.StatusCode != 200 {
-		return false, fmt.Errorf("got %d from %q %s", resp.StatusCode, endpoint, body)
+		return false, fmt.Errorf(
+			"got %d from %q %s", resp.StatusCode, stripToken(endpoint.String()), body)
 	}
 
 	if err := json.Unmarshal(body, &orgs); err != nil {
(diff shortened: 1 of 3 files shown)


nickmiller-wf:
For the record, I'm not tied to the name `stripToken` - naming these little helper functions can be tough sometimes. I'm happy to change it if something else seems more appropriate.

jehiah:
perfect. and a test for bonus points!

Can you squash down to a single commit?

nickmiller-wf:
🎉

nickmiller-wf:
@jehiah When should we expect a new release to be cut? Was having some trouble trying to build locally

PR #277, opened 2016-07-13
Roles header with refresh configuration

tlots:
This enhancement provides a new configuration to send roles as a header with an implementation for the Github provider. 

Roles are not stored in the cookie (per https://github.com/bitly/oauth2_proxy/issues/174#issuecomment-1578273584) but in the case of a restart of the oauth2_proxy application, the roles are retrieved. 

Additionally, metered updates of roles can be enabled with the cookie-refresh configuration to apply privilege changes to the current session.

PR #295, opened 2016-08-15
Interop for github provider and EmailDomains

donaldguy:
People (like my coworkers) often use the same github account for both work and other purposes. As such they end up having multiple email addresses set up for the same account

When using the pass-basic-auth, the X-Forwarded-Email and X-Forwarded-User end up passing their primary email and prefix thereof, which due to githubs configurability of notifications, may or may not be their work email. For systems which use the values of these headers to autovivify downstream users, confusion is introduced into the naming schemes, and in especially bad cases you could accidentally end up sending confidential notifications out of band.

While obviously there are more policy-level solutions, this is my proposal at a technical solution.

I suspect there may be a better way to achieve it, but this is a first pass; Let me know what you think.  Thanks :D

### DIFF
--- a/options.go
+++ b/options.go
@@ -212,10 +212,11 @@ func (o *Options) Validate() error {
 
 func parseProviderInfo(o *Options, msgs []string) []string {
 	p := &providers.ProviderData{
-		Scope:          o.Scope,
-		ClientID:       o.ClientID,
-		ClientSecret:   o.ClientSecret,
-		ApprovalPrompt: o.ApprovalPrompt,
+		Scope:                 o.Scope,
+		ClientID:              o.ClientID,
+		ClientSecret:          o.ClientSecret,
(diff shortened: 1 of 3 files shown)


donaldguy:
Oops accidentally hit submit without more info. will edit.

donaldguy:
To answer the ~obvious question of why use github if I care about email validation, I am using this cherry-picked on top of #277 to front Rundeck and as far as I understand, github is my best source of the roles data Rundeck needs to function

tlots:
Hi Donald, 

I like this enhancement, it's immediately useful for the Github provider. 👍 

A related case for reference: In a recent implementation we couldn't rely on all users having the same email domain (contractors, for example). So we are passing the user's Github ID as a unique identifier in lieu of the mailbox name - https://github.com/bitly/oauth2_proxy/pull/276

John

PR #305, opened 2016-09-17
Discovery backends by listening to kubernetes services.

pedro-r-marques:
This patch has the ability for the oauth2_proxy to automatically discovery kubernetes services.

In my current job, we use GKE for all of our services, which typically expose debug ports that we want to maintain under access control. In order to limit the access we use an oauth2-proxy with external access (via GCE LoadBalancer) that demuxes into multiple internal services. So far we have been modifying a configuration file when services are launched but that introduces extra manual work.

The following is an example configuration for a kubernetes service:

```
apiVersion: v1
kind: Service
metadata:
  name: embedding-svc
  namespace: roque
  annotations:
    oauth2-proxy/path: /embedding_svc/
spec:
  ports:
    - name: default
      port: 8080
      targetPort: 8080
  selector:
    run: embedding-dvc
```

The annotation ```oauth2-proxy/path``` causes the discovery module to create URL forwarding rule.
As far as i understand it, the oauth2-proxy operation is the same for dynamically learnt paths or paths discovered via the configuration file.

Please consider incorporating this functionality. I apologize in advance for not being aware of the coding conventions used in the project. I'm willing to refactor the code to comply with them but would appreciate guidance.

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -7,3 +7,4 @@ gopkg.in/fsnotify.v1                     v1.2.0
 golang.org/x/oauth2                      04e1573abc896e70388bd387a69753c378d46466
 golang.org/x/oauth2/google               04e1573abc896e70388bd387a69753c378d46466
 google.golang.org/api/admin/directory/v1 a5c3e2a4792aff40e59840d9ecdff0542a202a80
+k8s.io/client-go		 93fcd402979cfad8a7151f96e016416947c6a3cb
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -1,0 +1,28 @@
+package discovery
+
+import (
+	"net/http"
+	"net/url"
+)
+
+// Discovery defines the interface for backend discovery
+type Discovery interface {
+	NewServeMux(http.Handler) http.Handler
+}
+
+type nullDiscovery struct {
+}
+
+func (n *nullDiscovery) NewServeMux(mux http.Handler) http.Handler {
+	return mux
+}
+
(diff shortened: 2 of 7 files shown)


ploxiln:
This is pretty cool, and I'm sure some people will find it useful. But I think oauth2_proxy is trying to avoid further scope creep. I think it wants to focus on authentication and related options, and leave "interesting" upstream routing (and transformation) to another chained proxy.

For example, proxying to a private S3 bucket was an occasional but persistent request. Jehiah saw some utility in that himself, so he implemented it in a way that is most appropriate for use with oauth2_proxy: as a separate utility service. https://github.com/jehiah/private_s3_httpd

Also, this project is clearly a bit under-maintained. I certainly don't maintain it - I don't currently use it. I used it a lot in the past, soon after its inception, so I occasionally try to answer a question. For many upstream-routing questions, I suggest putting nginx upstream of oauth2_proxy. You can even put the same nginx process both downstream and upstream of oauth2_proxy for maximum flexibility, which I've done in the past. (And now there's "auth_request" as a simpler alternative to that.) That won't do anything like what you've implemented here, just another example of a typical use case. And I think this project should try to avoid answering nginx config questions (as much as I've taken the bait and attempted to answer them in the past), just focusing on the authentication, and letting the user get help for any upstream problems at a more appropriate venue. You could provide that appropriate venue for a kubernetes-debug-proxy 😁

pedro-r-marques:
https://github.com/pedro-r-marques/k8s-service-proxy

PR #323, opened 2016-11-18
Default to online access

stellacotton:
We don't need offline access for Zipkin, so lets restrict it to online access only.


stellacotton:
Sorry about that- didn't mean to make a PR against Bitly's version. :D

PR #328, opened 2016-11-29
Add custom header and title options.

mayfield:
Complements to the `-footer` argument for a header and sign in page
title.


mayfield:
This project seems to be dead.

jehiah:
If you want custom templates you can use the `-custom-templates-dir` option.

PR #329, opened 2016-12-02
Proxy rd argument to sign in form, fallback to RequestUri if not set

ean:
The hidden rd input field on the SignIn page ended up being /oauth2/signin?rd=/.... instead of /... when using nginx auth_request.


ploxiln:
That code in `SignInPage()` is now:

```go
	redirect_url := req.URL.RequestURI()
	if req.Header.Get("X-Auth-Request-Redirect") != "" {
		redirect_url = req.Header.Get("X-Auth-Request-Redirect")
	}
	if redirect_url == p.SignInPath {
		redirect_url = "/"
	}
```

So I think this pull request is no longer needed.

Issue #205, opened 2016-02-08
X-Forwarded-User should be configurable independent of -pass-basic-auth option


omazhary:
We are using oauth2_proxy in front of [grafana](http://docs.grafana.org/installation/configuration/#authproxy) with its "auth.proxy" configuration option. This relies on X-Forwarded-User header to be sent.
However if we use option `-pass-basic-auth`, we not only get X-Forwarded-User but also Authorization: Basic with the base64-encoded username and an empty password. This will fail the login to grafana.

Instead, we want to be able to only send X-Forwarded-User (and X-Forwarded-Email) independent of BASIC auth.
Will provide a patch soon.

johnrengelman:
+1 Just smashed into this exact issue as well. Changes LGMT.

johnrengelman:
Was just trying this patch out, but there still doesn't seem to be a way to log in to Grafana with Basic Auth once this proxy is in front of it (maybe that makes sense), but I'm trying to get at the built in admin account, but I always get redirected back to my oauth login.

omazhary:
Hi John, our scenario only relied on OAuth authentication, that's why we chose this proxy. It performs the authentication with grafana in the background and then logs you in directly to the dashboards. Making the regular grafana login page accessible would make the login process a little confusing.

Regarding the built-in admin account, you should be able to login using the UAA admin, as long as it is the exact same as the one in grafana. That is a bug, but has nothing to do with the proxy. This is basically because grafana calls its default administrator "admin", as well as the UAA. And, since grafana only matches the usernames, you can use any "admin" user to login to the grafana super-user. Ideally, this user should be changed in grafana, by being assigned a different username, either as a default, or by whoever is responsible for installing the system. Then, the new user would have to be created in cloud foundry's UAA as well.

johnrengelman:
Got it.

Definitely a bit annoying. Not really an issue with this tool, but was unexpected. Would be nice if Grafana had a property to promote specific accounts to admin on creation. Anyway, I worked around it, but bypassing the proxy, then logging in with the admin account and promoting my OAuth account to be an admin.

Thanks for the pointers.

omazhary:
No problem ;)

grahamrhay:
I have the same problem in front of Jenkins, the basic auth header causes the reverse proxy plugin to error, but I need the X headers. It would be great if this was merged.

zlatkoc:
Same issue with Jenkins as @grahamrhay reported. 

[...]

PR #339, opened 2017-01-24
Allow to pass user headers only

jsievers:
* This fixes https://github.com/bitly/oauth2_proxy/issues/205
* Add new boolean option -pass-user-headers
  to control whether X-Forwarded-User and X-Forwarded-Email
  headers will be set (as opposed to HTTP BASIC auth)
* This is required e.g. for grafana [1] where
  X-Forwarded-User is needed but HTTP BASIC auth fails
  (password is not known and must not be known in this scenario)
* Keep behaviour of PassBasicAuth unchanged for compatibility

[1] http://docs.grafana.org/installation/configuration/#authproxy


jsievers:
@jehiah thanks for merging!

PR #346, opened 2017-02-23
Oversize Cookie Alert

bdwyertech:
Cookies cannot be larger than 4kb

It took me a long while to debug this, no one else should ever have to.

This issue showed face whilst attempting to add `refresh_token` support to the Azure provider, which uses pretty large tokens.

With all the padding and what not, I was hitting 4264 bytes, which was simply triggering the alert:
`Cookie "oauth2_proxy" not present`

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -255,6 +255,10 @@ func (p *OAuthProxy) MakeCookie(req *http.Request, value string, expiration time
 
 	if value != "" {
 		value = cookie.SignedValue(p.CookieSeed, p.CookieName, value, now)
+		if len(value) > 4096 {
+			// Cookies cannot be larger than 4kb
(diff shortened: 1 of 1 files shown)


bdwyertech:
@thenewwazoo you might want to take this into consideration in your Azure implementation.  

Even without any groups (like you've added to the cookie), the combination of `access_token` and `refresh_token` alone make the resulting cookie too large.

thenewwazoo:
Yeah, exploding cookie size was something I expected given that there are instances when a user in our org can be a member of literally hundreds of groups. The various filtering mechanisms were an attempt to corral that. I haven't yet encountered a scenario where cookie sizes get too big without groups, but I also haven't been tracking their actual size.

I'm tempted to bust group information out into an entirely separate cookie.

Issue #335, opened 2017-01-09
Need Azure Active Directory group claims support


thenewwazoo:
Azure AD has the ability to provide [group membership claims](https://blogs.technet.microsoft.com/enterprisemobility/2014/12/18/azure-active-directory-now-with-group-claims-and-application-roles/) in the JWT (or, if there are too many groups, a `group.link` value that the application can use to obtain a paged list of groups).

I'd like to be able to pass this data along to the proxied application in the `X-Forwarded-Groups` header.

This looks fairly straightforward. Are there any caveats I should keep in mind when adding this?

thenewwazoo:
It also looks like AAD supports [roles claims](http://www.dushyantgill.com/blog/2014/12/10/roles-based-access-control-in-cloud-applications-using-azure-ad/) as well. If it's straightforward, I may add explicit support for that as well.

thenewwazoo:
#347 adds this support.

PR #347, opened 2017-03-14
Improve support for AAD Auth proxying w/ group authz support

thenewwazoo:
This PR improves support for backing oauth2_proxy with Azure Active Directory, along with a few other changes. We've been using this internally for lightly-loaded services for a couple of weeks, and are ready to open contribute it back upstream. Thanks for the great base upon which to build!

**BREAKING CHANGES**:
* In order to generalize group support, the Google `google-groups` field was renamed to `permit-groups`.
* Raw cookie value construction was changed to delineate fields using the `:` character instead of the `|` character in order to adopt conventional construction of the `X-Forwarded-Groups` header (usually `foo|bar|baz`).
* The `provider` option is now mandatory.

Added functionality:
* Azure support now includes group support, including passing an `X-Forwarded-Groups` header, filtering groups (in order to limit cookie size), and restricting access based on group string matches.
* Added a Vagrantfile and simple HTTP request reflector for testing


Esfera5:
ETA?

thenewwazoo:
Sorry, I don't follow. We've been using this internally for a bit now. Does it need something to be ready for merge?

bdwyertech:
FYI, I went live with this and it turns out that having `prompt=consent` as a query param for initial authorization does not work for non-admin users.  I had to revert back to the mainline version here without the `prompt` query param.

thenewwazoo:
@bdwyertech, thanks for the report. I'm unfortunately no longer in a position to test this, but did you have any success with [different values](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols-oauth-code#request-an-authorization-code) for `prompt`? Thinking about it, it is probably advisable to remove it altogether, since I don't believe it's necessary to require the consent prompt every time.

I'm surprised to hear you had that result. I don't believe that everyone was an admin in my former team. Hopefully it's a simple fix. Do let @pasoroki know if you're able to do any testing, as he may be interested in updating this PR. :)

pasoroki:
I will definitely update the PR.
We have a ticket for one person who didn't get 2FA request when he tried to authorize. That might be related. I'll try first to remove `prompt` at all from the request

pasoroki:
This PR is updated:
 - merged latest changes from the origin
 - added empty `prompt` option by default
 - fixed issue with HTTP 500 error if `state` option were not set

The `test.sh` throw lots of errors, they are unrelated to our changes.

providers/google_test.go
```diff
@@ -110,16 +110,16 @@ func TestGoogleProviderValidateGroup(t *testing.T) {
 	p.GroupValidator = func(email string) bool {
 		return email == "michael.bland@gsa.gov"
 	}
-	assert.Equal(t, true, p.ValidateGroup("michael.bland@gsa.gov"))
+	assert.Equal(t, true, p.ValidateGroup(&SessionState{Email:"michael.bland@gsa.gov"}))
```
ploxiln: `test.sh` is currently stopping due "go fmt" not liking this line which you changed

ploxiln:
The test panic is:

```
--- FAIL: TestAuthOnlyEndpointSetXAuthRequestHeaders (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
...
github.com/bitly/oauth2_proxy.NewOAuthProxy(0xc4201f9800, 0xc42021e070, 0xc42021e060)
	/home/travis/gopath/src/github.com/bitly/oauth2_proxy/oauthproxy.go:163 +0x590
github.com/bitly/oauth2_proxy.TestAuthOnlyEndpointSetXAuthRequestHeaders(0xc4201f66c0)
	/home/travis/gopath/src/github.com/bitly/oauth2_proxy/oauthproxy_test.go:628 +0x1fe
```

I think I that's on this line:

```go
log.Printf("OAuthProxy configured for %s Client ID: %s", opts.provider.Data().ProviderName, opts.ClientID)
```

so `opts.provider.Data().ProviderName` is probably de-referencing a nil somewhere.

You are making a significant change to how providers work in this PR: making the provider no longer default to "google" but instead return `nil` (and an error) if no recognized provider is specified. So it is quite plausible that this provider-option-handling change is related to this test failure.

Command-line common-use-case compatibility-breaking changes should not be bundled with other changes like group handling enhancements, and is probably enough to prevent this from being merged. On the other hand, while this project/repo is not "dead", it is not clear what might be merged next, or in how many weeks or months. It looks like any of the changes you want to get in might be a bit too big to get into the next round, not really because they're "bad", but just because review capacity of the maintainers is low. (I'm not a maintainer.)

If this PR is just useful as a common resource for multiple users of this variant/functionality, that's quite fine, you can ignore me :)

pasoroki:
@ploxiln you were right.
We removed default provider and were defining provider for each option set in every test.
New tests didn't have that. I removed back Google as a default provider

ploxiln:
fwiw, I don't think the more strict provider option is a bad idea, just unrelated :)

(I would still have it default to google, but right now it accepts any unrecognized string as google, so that could be cleaned up - in a separate PR)

johnarnold:
@talem @jediah   This is really useful, any issues blocking merge?

johnarnold:
"In order to generalize group support, the Google google-groups field was renamed to permit-groups."

consider leaving the google-groups param along with permit-groups,for backwards compatibility, and set 
permit-groups = google-groups   if permit-groups is none, and google-groups is not none.

iksaif:
Agreed with @johnarnold, having this merged would be great.

thefatalist:
I'll update PR next week, but don't think anyone would merge it =(

iksaif:
why is that ? @jehiah tagged the associated issue (#335) as enhancement, so I don't think authors would be against this.

thefatalist:
Well, I hope you are right @iksaif

providers/session_state.go
```diff
@@ -71,19 +71,41 @@ func (s *SessionState) EncryptedString(c *cookie.Cipher) (string, error) {
 			return "", err
 		}
 	}
-	return fmt.Sprintf("%s:%s:%d:%s:%s", s.userOrEmail(), a, s.ExpiresOn.Unix(), r, s.Groups), nil
+
+	encoded_groups := base64.StdEncoding.EncodeToString([]byte(s.Groups))
+
+	return fmt.Sprintf("%s|%s|%d|%s|%s", s.accountInfo(), a, s.ExpiresOn.Unix(), r, encoded_groups), nil
```
iksaif, line 78: maybe addind some base64 encoded json would be better ?

thefatalist: This is original flow. I very much like the idea, but I don't really want now to do more global changes.
P.S. Looks like tests are failing because Travis is running out of space:
```
2017/12/11 12:36:19 watcher.go:65: failed to add /tmp/test_auth_emails_834915528 to watcher: no space left on device
exit status 1
FAIL	github.com/bitly/oauth2_proxy	0.111s
```

iksaif: Ok, makes sense.

maybe try to re-trigger the test ?

providers/provider_default.go
```diff
@@ -36,16 +39,18 @@ func (p *ProviderData) Redeem(redirectURL, code string) (s *SessionState, err er
...
 
 	var resp *http.Response
-	resp, err = http.DefaultClient.Do(req)
-	if err != nil {
-		return nil, err
-	}
+	resp, c_err := http.DefaultClient.Do(req)
 	var body []byte
-	body, err = ioutil.ReadAll(resp.Body)
+	body, b_err := ioutil.ReadAll(resp.Body)
```
combor, line 47: `ioutil.ReadAll` will panic if the `resp` is `<nil>`

pingles: @combor because `c_err` isn't being handled?

combor: Correct. It's handled below but it's to late :)

pasoroki: Good catch, thanks guys!

providers/azure.go
```diff
@@ -8,34 +8,40 @@ import (
...
-			Host:   "graph.windows.net",
+			Host:   "graph.microsoft.com",
 		}
 	}
 	if p.Scope == "" {
 		p.Scope = "openid"
 	}
 
+	if p.ApprovalPrompt == "force" {
+		p.ApprovalPrompt = "consent"
```
combor, line 42: The default approval prompt is set to `force` so this will be always set to `consent` which brakes it for non admin users. I had to set `-approval-prompt=""` to make it work. This was mentioned already in https://github.com/bitly/oauth2_proxy/pull/347#issuecomment-305943432

combor:
The `-pass-groups=true` parameter will only work if `-pass-access-token` is set. The reason is that [groups are saved in the session cookie](https://github.com/bitly/oauth2_proxy/pull/347/files#diff-e8319279ea5bcf2bfe02dda370ee30ecR75) which is created inside `EncryptedString` method that requires `c *cookie.Cipher`. This is only set when: https://github.com/bitly/oauth2_proxy/blob/731fa9f8e00b294ff1bf4a687e63d2ecdd9a4a50/oauthproxy.go#L166

I think this needs some explanation in docs or change of the condition so if `pass-groups` is set than the `pass-access-token` is a required flag.

pasoroki:
@combor PassAccessToken is setup by default
https://github.com/bitly/oauth2_proxy/blob/master/options.go#L111
I marked in README that its default value is `false`

providers/azure.go
pasoroki: In mentioned comment we had a discussion that there is no way to use `none` as an approval prompt
I changed it to "" by default now

pasoroki:
Thank you @jan-schulz-k24  that is very helpful
Now I need to figure why golang 1.9 is timing out

providers/provider_default.go
```diff
@@ -36,15 +39,17 @@ func (p *ProviderData) Redeem(redirectURL, code string) (s *SessionState, err er
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
 	var resp *http.Response
-	resp, err = http.DefaultClient.Do(req)
-	if err != nil {
-		return nil, err
+	resp, c_err := http.DefaultClient.Do(req)
+	if c_err != nil {
+		return
```
combor: If you don't pass `c_err` to the return statement it will exit with `err` as `<nil>` because by default the function can only return those named variables`(s *SessionState, err error)`.

pasoroki:
Hey guys, any updates so far?
Anything else I can do?

nduijvelshoff:
Hejj all, what us the current status of this? Can't wait to use it. 👍

combor:
We've been using it for couple of months now from our fork and it works ok

nduijvelshoff:
@combor When I clown the fork, the build fails on my opensuse server...

nduijvelshoff:
Clone*

pingles:
@nduijvelshoff could you give us the errors you're getting/build output and we'll see if we can debug the problem

combor:
@nduijvelshoff You can also use this container: `quay.io/uswitch/oauth2_proxy`

nduijvelshoff:
@combor This won't suite for me since I'm reusing the Kibana docker image.
@pingles Hereby my error:
./oauthproxy.go:569: session.IDToken undefined (type *providers.SessionState has no field or method IDToken)
./oauthproxy.go:570: session.IDToken undefined (type *providers.SessionState has no field or method IDToken)
./oauthproxy.go:571: p.provider.GetGroups undefined (type providers.Provider has no field or method GetGroups)
./oauthproxy.go:576: session.Groups undefined (type *providers.SessionState has no field or method Groups)
./oauthproxy.go:585: cannot use session (type *providers.SessionState) as type string in argument to p.provider.ValidateGroup
./oauthproxy.go:702: session.Groups undefined (type *providers.SessionState has no field or method Groups)
./oauthproxy.go:703: session.Groups undefined (type *providers.SessionState has no field or method Groups)
./options.go:282: assignment count mismatch: 2 = 1
./options.go:291: p.SetGroupRestriction undefined (type *providers.AzureProvider has no field or method SetGroupRestriction)

pingles:
Wow, super strange. I've take a clean clone on a new laptop and can build with:

```
$ go get -v -t ./...
$ CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/proxy
```

I'd say maybe its a dependency versioning issue but then the options go assignment error suggests something else is wrong.

nduijvelshoff:
@pingles going to check it out tonight.... can you share the complied bin with me? 👼

combor:
You can find it in our container.

nduijvelshoff:
@pingles are you sure we are talking about the same repo? https://github.com/outlook/oauth2_proxy.git
Tested it on Suse and Ubuntu (Go 1.10).

pingles:
@nduijvelshoff ah, yeah @combor and i were talking about our changes in our fork. code is at https://github.com/uswitch/oauth2_proxy

nduijvelshoff:
@pingles @combor I get it, I see it also supports the AAD groups... Going with that one. Thanks!

nduijvelshoff:
@pingles is it possible to add -basic-auth-username to overwrite the "x-user" to combine it with:  "-basic-auth-password string: the password to set when passing the HTTP Basic Auth header"

kerma:
Any plans to support groups in `/auth` request as well (for Nginx `auth_request`)? It took me some hours to figure out why I don't see any groups there 🤦‍♂️

pasoroki:
Sorry @kerma I don't work on that

kerma:
@pasoroki would you accept a PR with that functionality ?

pasoroki:
@kerma sure, just don't forget to cover code with tests

pasoroki:
@kerma sure, just don't forget to cover code with tests

jcrowthe:
When might this be merged? Groups support is really the only path forward when it comes to Azure provider support for companies larger than ~10 individuals.

lukasmrtvy:
Merge? We really need this feature. Thanks

mkoch1:
This looks like exactly what we need - any idea when it can get merged (and released)?

dekimsey:
I've played with this PR and haven't been able to get Azure Roles or Groups passed through. Does anyone have docs on the Azure set-up side of things?

lukasmrtvy:
@dekimsey you will have to use v2.0 api endpoint and scope: profile I guess

dekimsey:
@lukasmrtvy Thanks, that made me look into this deeper. I haven't been able to see/get roles but I did get groups working. It was a number of things some PEBKAC others permissions issues with the Graph API (user_list_memberof)[https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/user_list_memberof].

WRT to the PR itself, I did find that a large group list will cause it to go into loop hitting the azure cookie /oauth2/callback repeatedly as the Cookie maximum size is exceeded and unable to save in the browser. To whom whoever decides to merge this or take this over, might be worth catching that situation rather than infinite looping. Also, the FilterGroups parameter cannot be specified multiple times which makes filtering tricky.

mariusfylling:
This looks like exactly what we need - any idea when it can get merged?

lukasmrtvy:
@mariusfylling this is dead project. Keycloak and Keycloak Gatekeeper can do the magic for You. 
Keycloak Gatekeeper itself can replace this oauth2 proxy, but there is a small problem: https://issues.jboss.org/browse/KEYCLOAK-8728

mariusfylling:
> @mariusfylling this is dead project. Keycloak and Keycloak Gatekeeper can do the magic for You.
> Keycloak Gatekeeper itself can replace this oauth2 proxy, but there is a small problem: https://issues.jboss.org/browse/KEYCLOAK-8728

Thank you, will look into this.

martin-loetzsch:
Fyi: there is an active discussion about forking this project here: https://github.com/bitly/oauth2_proxy/issues/628

PR #350, opened 2017-03-21
Allow to skip SSL verification of Oauth2 provider.

Gui13:
This is useful when testing oauth2_proxy in a dev environment, where the local proxy might not have valid certificates.


jehiah:
@Gui13 can you clarify your use case? Even if you run oauth2 proxy with self-generated certificates. the requests it makes are against the OAuth providers and should always be validated.

Gui13:
Hi @jehiah, I'm using this option to validate both oauth proxy connection to our sso locally (running my own instance of an SSO service on my machine, which means https://127.0.01) and for unit tests, where we deploy our sso in a container with an internal domain name that has no valid certificate.

I understand your doubts about this feature because it could be dangerous for a production environment, but for internal unit testing, it is often required. Maybe I could add some traces and comments that this shouldn't be used in production?

PR #358, opened 2017-03-28
Perform recursive lookups for group members that are themselves groups

mtak:
In Google groups it is possible to make groups members of other groups. If A is
a member of group G and group G is a member of group H, then A should also be
considered a member of group H.

This change will append any groups found when listing group members to the
array of groups to iterate through, so that eventually all top-level and nested
groups will be used to check membership for the user. It is not clear whether
cyclic group membership is permitted by Google but it is being checked for just
in case.

### DIFF
--- a/providers/google.go
+++ b/providers/google.go
@@ -194,7 +194,10 @@ func userInGroup(service *admin.Service, groups []string, email string) bool {
 	id := user.Id
 	custID := user.CustomerId
 
-	for _, group := range groups {
+	groups = groups[:]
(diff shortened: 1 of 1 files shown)


snomann:
Why is this not merged?

framillien:
Hello, looking around for groups handling, I found this MR also reworking this part: https://github.com/bitly/oauth2_proxy/pull/500

Looks like this may fix this problem, with simplified code: groups are filtered by user email, server side (scheduled for 2.3).

Issue #321, opened 2016-11-18
Cross-Site Request Forgery (CSRF)


permyakovsv:
missing state parameter in google oauth login

lvh:
This has been fixed, see #336.

PR #360, opened 2017-03-29
CSRF protection for OAuth flow.

### LINKED ISSUE #336: csrf validation
The OAuth2 state parameter is used as redirect target URL (it could be like a "next" maybe instead?). The state parameter should be secure random thingy we can check in the callback handler:

https://tools.ietf.org/html/rfc6749#section-4.1.1

> The parameter SHOULD be used for preventing
         cross-site request forgery

Also: http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html

Pls fix this. Thx.

jehiah:
This implements CSRF protection for the initiation and termination of the OAuth flow. (thanks @arnottcr and @therook for the contribution).

Fixes #336, #321

### DIFF
--- a/cookie/nonce.go
+++ b/cookie/nonce.go
@@ -1,0 +1,16 @@
+package cookie
+
+import (
+	"crypto/rand"
+	"fmt"
+)
+
+func Nonce() (nonce string, err error) {
+	b := make([]byte, 16)
+	_, err = rand.Read(b)
+	if err != nil {
+		return
+	}
+	nonce = fmt.Sprintf("%x", b)
+	return
+}
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -37,6 +37,7 @@ var SignatureHeaders []string = []string{
 type OAuthProxy struct {
 	CookieSeed     string
 	CookieName     string
+	CSRFCookieName string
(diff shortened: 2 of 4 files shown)


colemickens:
@jehiah / @arnottcr Can you help me understand specifically why this PR was submitted?

I'm using oauth2 proxy at `https://auth.mydomain.tld` to protect a resource `https://super-secret.mydomain.tld`.

RE: this specific function: https://github.com/bitly/oauth2_proxy/commit/55085d9697962668fd4e43e8e4644144fe83cd93#diff-6fd8df33d9f8086bc31e28c375fbc0abL387

AFAICT, this code means that hitting `https://auth.mydomain.tld/oauth2/start?rd=https://super-secret.mydomain.tld` will never work, because the redirect is always reset to `/`.

Does it ever make sense to relax this?

colemickens:
I may not have linked the exact right location in code, there are a couple functions that "filter" the redirect down to `'/'`.

It seems like this was done for security reasons. Would it be sufficient to accept a list of domains to respect? (For example, I am already having oauth2proxy set the cookie domain to mydomain.tld, can I also have a `--redirect-whitelist=mydomain.tld` or some such?

Issue #351, opened 2017-03-22
Config file and command line parameters do not match up - v2.2.0-alpha


iyerusad:
I built from source
```
# Build oauth2_proxy
$ export GOPATH=$HOME/go
$ sudo go get github.com/bitly/oauth2_proxy
$ sudo cp ~/go/bin/oauth2_proxy /usr/bin/
$ oauth2_proxy -version
oauth2_proxy v2.2.0-alpha (built with go1.6.2)
```

Working **cmdline** syntax:
```
$ oauth2_proxy \
-provider="azure" \
-client-id="blah-blah \
-client-secret="supersecret" \
-cookie-secret="randomsecret" \
-cookie-secure=false \
-email-domain="company.com" \
-upstream="http://127.0.0.1:8080/"
```

Working **cfg file** syntax:
```
provider="azure"
client_id="blah-blah"
client_secret="supersecret"
cookie_secret="randomsecret"
cookie-secure="false"
email_domains="company.com"
upstreams="http://127.0.0.1:8080/"
```

cookie_secret is expected in config file, cookie-secret is expected in command line
email_domains is expected in config file, email-domain is expected in command line

so on and so forth

jehiah:
It's expected that command line arguments are `-with-dashes` and config file keys are `with_underscores` so `cookie-secure` in the config file seems like the only one from your list that's out of place.

I think the singular vs plural for email_domains is expected because they have different ways to specify multiple values (but please confirm).

iyerusad:
Ah, I see. Is there some area where that disparity is documented? I dug through some of the basic documentation but might of missed it. It felt somewhat like a bug based on what the error message - the error message says do A syntax but B syntax is what is really desired.

jehiah:
If you have thoughts on specific documentation updates that would have been helpful to you, i'm more than happy to review a PR for that.

PR #363, opened 2017-03-30
Update README.md

iyerusad:
Added blurb about difference between config and command line options with an example config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,22 @@ Usage of oauth2_proxy:
   -version=false: print version string
 ```
 
+### Config File Options
+
+The config file options are not always the same as command line options. Details on config file options: (information needed)
+
+Example config file:
+```
+provider="azure"
+client_id="$clientID"
+client_secret="$appsecret"
+cookie_secret="$cookiesecret"
+cookie-secure="false"
+email_domains="$emaildomain"
+upstreams="http://127.0.0.1:8080/"
+skip-provider-button="true"
+```
+
 See below for provider specific options
 
 ### Upstreams Configuration


iyerusad:
this pull request fixes #351

jehiah:
@iyerusad I'd rather fix the spots where it's inconsistent so there is nothing to document.

options.go
```diff
@@ -19,60 +19,60 @@ import (
 // Configuration Options that can be set by Command Line Flag, or Config File
 type Options struct {
 	ProxyPrefix  string `flag:"proxy-prefix" cfg:"proxy-prefix"`
```
ploxiln: I could be wrong, but I think what was requested was for this one to change, not the rest.

I don't think we want to change most of the options and break compatibility with all config files. Rather, I think we just want it to be consistent: always `-` for flag, always `_` for config-file variable (and the rest of the letters consistent, if any happen to not be). This is a familiar convention for unix programs, believe it or not :)

options.go
```diff
@@ -119,7 +119,7 @@ func parseURL(to_parse string, urltype string, msgs []string) (*url.URL, []strin
 
 func (o *Options) Validate() error {
 	msgs := make([]string, 0)
-	if len(o.Upstreams) < 1 {
```
iyerusad: Oops I see why this broke.

options.go
iyerusad: I don't mind either way, as long as there is some sanity in consistency. 

What say you all:
    **Option A)** cmdline flags as "variable-name", config file values as "variable_name"
    **Option B)** cmdline flags and config file values as "variable-name"
    **Option C)** cmdline flags and config file values as "variable_name"
    **Option D)** ???

ploxiln: Option A

iyerusad:
Okay updated changes to go with:
**Option A)** cmdline flags as "variable-name", config file values as "variable_name"

options.go
```diff
@@ -48,7 +48,7 @@ type Options struct {
 	CookieSecure   bool          `flag:"cookie-secure" cfg:"cookie_secure"`
 	CookieHttpOnly bool          `flag:"cookie-httponly" cfg:"cookie_httponly"`
 
-	Upstreams             []string `flag:"upstream" cfg:"upstreams"`
+	Upstreams             []string `flag:"upstream" cfg:"upstream"`
```
ploxiln: The reason why some of these were singular for the flag but plural for the cfg, was that when using the flag you use multiple times:

```
oauth2_proxy --upstream=http://127.0.0.1:4190/A --upstream=http://127.0.0.1:4191/B
```

but for the cfg you pass them all together:

```
upstreams = [
    "http://127.0.0.1:4190/A",
    "http://127.0.0.1:4191/B"
]
```

... so there was a logic to it. (Also, "upstreams" and "email_domains" are the most common options, so backward compatibility is probably most important for them.)

Thanks for pointing out these differences. We'll have to wait for @jehiah to have time to re-review.

PR #365, opened 2017-04-01
Fix url parse error

travisofthenorth:
Fixes this error which I was running into locally:
`2017/04/01 14:59:25 http.go:29: FATAL: could not parse "127.0.0.1:4180": parse 127.0.0.1:4180: first path segment in URL cannot contain colon`

Update: the above error happens as of go 1.8. This PR removes parsing via `url.Parse` and parses the scheme manually.

### DIFF
--- a/http.go
+++ b/http.go
@@ -5,7 +5,6 @@ import (
 	"log"
 	"net"
 	"net/http"
-	"net/url"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


ploxiln:
The square brackets in the help text suggest that the "http://" is supposed to be optional. And, if you test the v2.1 release binary, it works without "http://" specified. So, this is a bug elsewhere in oauth2_proxy, maybe due to a library update.

travisofthenorth:
@ploxiln I'm guessing the release binary was built with go 1.7? Looks like the behavior changed slightly in 1.8 (https://github.com/golang/go/issues/19297). Thoughts? I'm happy to update this PR or close it if 1.8 isn't meant to be supported.

ploxiln:
The previous release was built with go-1.6 (it's in the tarball name). We do intend to support go-1.8 but were not aware of this change to url parsing yet. Thanks for the report and the link to an explanation.

I think that `ServeHTTP()` should be changed to parse the optional scheme without `url.Parse()`. All it really does is remove the scheme from the option string if it is present, and specify "tcp" or "unix" as appropriate to `net.Listen()` (whereas `ServeHTTPS()` just passes the option string directly to `net.Listen()`, always "tcp").

travisofthenorth:
@ploxiln how does this look? Tested it manually with: 

- `tcp://127.0.0.1:4180`
- `http://127.0.0.1:4180`
- `//127.0.0.1:4180`
- `127.0.0.1:4180`
- `unix:///tmp/foo.sock`

ploxiln:
That looks OK. I think it could be simplified a bit more, since oauth2_proxy has never accepted `//127.0.0.1:4180` ("panic: unknown upstream protocol"). (I tested 2.1 release built with go-1.6 and current master built with go-1.8.)

Anyway, we'll have to wait for a review from a maintainer, I'm not really one of them :)

travisofthenorth:
Sounds good. Thanks for taking a look!

ploxiln:
ping @jehiah 
recap: default value for flag `--http-address` no longer works in go-1.8

PR #314, opened 2016-10-27
WIP: LetsEncrypt Support via rsc.io/letsencrypt

fortytw2:
Not finished yet, but opening a WIP PR keeps the work I'm doing on this open 

Allows for automatic fetching and validation of LetsEncrypt SSL certs with a few quick settings :) 

TODO: 
- [ ] Update documentation
- [ ] Really test that it works (haven't done yet)
- [ ] Ensure HTTP -> HTTPS redirects when using LetsEncrypt

This will close #313

Issue #313, opened 2016-10-26
autocert / letsencrypt integration


fortytw2:
Would be extremely useful for oauth2_proxy to automatically grab SSL certificates from LetsEncrypt if asked to. I'm willing to implement this, if the maintainers think it's a good / mergable idea :)

jehiah:
@fortytw2 yes! I'd be happy to see this land using `https://github.com/rsc/letsencrypt`

fortytw2:
@jehiah fantastic! I'll try to send a PR tomorrow :)

PR #367, opened 2017-04-05
Add Let's Encrypt Support

devoid:
Build on @fortytw2's work and add Let's Encrypt TLS certificate support. This PR replaces PR #314 and addresses #313.

- Switched to use golang.org/x/crypto/acme/autocert
- Added more options tests.
- Added documentation in README and CLI prompts.
- Tested and running as expected in my production environment.

_I couldn't figure out a way to append to that existing PR so I created a new one._
_Sorry if this isn't the right way to do it._

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -4,6 +4,7 @@ github.com/bitly/go-simplejson           da1a8928f709389522c8023062a3739f3b4af41
 github.com/mreiferson/go-options         77551d20752b54535462404ad9d877ebdb26e53d
 github.com/bmizerany/assert              e17e99893cb6509f428e1728281c2ad60a6b31e3
 gopkg.in/fsnotify.v1                     v1.2.0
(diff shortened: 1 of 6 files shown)


devoid:
@fortytw2 I'll admit that it is a bit pedantic, but the [proxy doesn't support SSLv3](https://github.com/bitly/oauth2_proxy/blob/master/http.go#L58).

devoid:
@jehiah Can you take a look at this and let me know if you'd like to see anything different?

gbolo:
@fortytw2 we should all stop saying _SSL_ and start calling it **TLS**; I will also admit that it is a bit pedantic =)

akarnani:
If this is all set, could you merge it into mainline so I can stop using a fork?  Thanks!

mattbailey:
Is `oauth2_proxy` still an active project?  I'd really like to see this feature in rather than doing orchestration gymnastics with certbot.

PR #372, opened 2017-04-10
Adding official Docker image

dstroot:
I know there is another pull request for this already.  This addresses the idea that the docker container could be/should be tied to the version I check out and it creates a **TINY** container - literally the smallest possible.


ploxiln:
It would be better to actually build from local source with local go. If this is done by the maintainers as they make a release, it will be equivalent to the release built by dist.sh. (The download link will break often because the latest-go-version-at-release is also in it.)

It would be better to not maintain a set of root CA certificates in this repo, particularly since it's not obvious where exactly they came from. A better option may be to download them (from e.g. a trusted Mozilla URL) in the build step of the Makefile.

dstroot:
I agree with you.  The problem is there is really no place I know online to get the .crt file (I pulled the one here from the latest Ubuntu).  Mozilla only publishes a file which must be "processed" first.

https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/

Does anyone know a good source?  Or, I could use this approach: https://github.com/broady/cacerts

ploxiln:
I think debian ca-certificates is fine. Getting it from a docker image that contains just the ca-certificates is a bit odd (but would work). Another option is to download directly from https://raw.githubusercontent.com/certifi/python-certifi/master/certifi/cacert.pem

dstroot:
I could use the dist.sh script to build the binary exactly as if was built otherwise, however it always builds the latest version.  Thoughts about an approach like this:

```sh
build:
	# build the binary
	cd .. && \
	./dist.sh
	
	# get binary
	# cd docker && \
	cp ../dist/*linux*.* binary.tar.gz && \
	tar xfz binary.tar.gz --strip-components 1 \
	rm -rf binary.tar.gz
	
	# get the ca certificates. We get them from Certifi is a carefully curated 
	# collection of Root Certificates for validating the trustworthiness of 
	# SSL certificates while verifying the identity of TLS hosts.
	# https://github.com/certifi/python-certifi
	wget https://raw.githubusercontent.com/certifi/python-certifi/master/certifi/cacert.pem -O ca-certificates.crt
	
	# build the docker image
	docker build -t $(DOCKER_NAME)/$(IMAGE_NAME):latest .
```

ploxiln:
That is roughly what I had in mind, thanks. As another example, `dist.sh` for nsq runs "docker build" itself: https://github.com/nsqio/nsq/blob/master/dist.sh

idntfy:
re ca-certificates, we use alpine and our dockerfile looks like this for pre-built binary which I think is pretty simple and lightweight:

```
FROM alpine
RUN apk add --no-cache --virtual=build-dependencies ca-certificates
COPY oauth2_proxy /run
CMD /run/oauth2_proxy
```

or, we used this one before we started using custom prebuilt binary:

```
FROM alpine

RUN apk add --no-cache --virtual=build-dependencies wget ca-certificates && \
    wget -P /tmp https://github.com/bitly/oauth2_proxy/releases/download/v2.1/oauth2_proxy-2.1.linux-amd64.go1.6.tar.gz && \
    tar -C /tmp -zxvf /tmp/oauth2_proxy-2.1.linux-amd64.go1.6.tar.gz && \
    mv /tmp/oauth2_proxy-2.1.linux-amd64.go1.6/oauth2_proxy /run/oauth2_proxy

CMD /run/oauth2_proxy
```

reedloden:
You could also use https://curl.haxx.se/ca/cacert.pem or https://raw.githubusercontent.com/bagder/ca-bundle/master/ca-bundle.crt (though, seems the latter is deprecated in favor of the former).

https://curl.haxx.se/docs/caextract.html has more information on this.

ploxiln:
Curl's website does look like a good source for such a CA bundle. It's probably equivalent to what is currently being fetched from the Certifi project.

Interestingly, that page says:

> The converted PEM file only contains the digital signatures for CAs. Several of those CAs have constraints in Firefox (and other browsers) to only be allowed for certain domains and other similar additional conditions. Those constraints are thus not brought along in this cacert file! 

... which makes me wonder how many such constrained domain-signing CAs are present, if they are similarly present in the Certifi bundle (probably?), if it would be practical to omit them ... but I think those are questions for another time, no need to answer them here.

julianvmodesto:
Bump! Is there anything blocking for this PR or way I could contribute?

Would love an official image at Docker Hub or Quay (as I mentioned in the duplicate #416) so I could `docker pull bitly/oauth2-proxy:2.2 && docker run ...`, and further set up an oauth2-proxy sidecar in a Kubernetes Pod.

buckhx:
Bump, https://hub.docker.com/r/a5huynh/oauth2_proxy/ seems to be the only up to date one, but it's a 250MB image

skwashd:
Docker 17.05 and up supports [multistage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/). This would allow you to use `dist.sh` to build the oauth2_proxy binary in the first stage and then copy it and anything else you need into a clean `FROM scratch` image in the second stage.

It would be good to add a user in the container so it doesn't run as root. Should there be an `EXPOSE` directive to make the service available on port 4180?

I have done this with [another golang project](https://github.com/skwashd/lilypad-docker/blob/master/connect/Dockerfile). Feel free to grab what you find useful from it.

boivie:
I published a pull request to create a Docker image using multi-stage build.

It's recommended to build it at Docker Cloud - automated on every push and every tag. You will need to set this up yourself using your account, so that it's really official. I can help out if you want step-by-step instructions.

adamdecaf:
Any update on this?

martin-loetzsch:

PR #376, opened 2017-04-19
Don't set the cookie domain to the host by default, as it breaks Cookie Prefixes

### LINKED ISSUE #352: Allow exclusion of Domain= in cookies
I've set the cookie-name to `__Host-oauth2` in an attempt to get [Cookie Prefixes](https://tools.ietf.org/html/draft-west-cookie-prefixes-05#section-1) ([__Host-](https://tools.ietf.org/html/draft-west-cookie-prefixes-05#section-3.2)) to work but I'm having some issues getting Chrome to accept the cookie.

Section 3.2 forbids setting Domain= but I can't seem to find a vanilla way to exclude Domain= from the cookie oauth2_proxy sends -- could a "none" option be added for -cookie-domain?

reedloden:
The Cookie Prefixes spec disallows the use of the `domain` attribute in cookies
if the `__Host-` prefix is used
(https://tools.ietf.org/html/draft-ietf-httpbis-cookie-prefixes-00#section-3.2).

There's no need to set it to the host by default, so make it optional. If it is
set to a non-empty value, still output a warning if it is not a suffix of the
host, as that's likely not wanted.

Fixes #352.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Usage of oauth2_proxy:
   -client-id string: the OAuth Client ID: ie: "123456.apps.googleusercontent.com"
   -client-secret string: the OAuth Client Secret
   -config string: path to config file
-  -cookie-domain string: an optional cookie domain to force cookies to (ie: .yourcompany.com)*
+  -cookie-domain string: an optional cookie domain to force cookies to (ie: .yourcompany.com)
   -cookie-expire duration: expire timeframe for cookie (default 168h0m0s)
   -cookie-httponly: set HttpOnly cookie flag (default true)
   -cookie-name string: the name of the cookie that the oauth_proxy creates (default "_oauth2_proxy")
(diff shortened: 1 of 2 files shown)


ploxiln:
After some reading, I agree this is a good idea, and this implementation looks good to me. It'll need a good little explanation in the changelog of the next release.

According to both https://en.wikipedia.org/wiki/HTTP_cookie#Domain_and_Path and http://erik.io/blog/2014/03/04/definitive-guide-to-cookie-domains/, specifying the Domain in the cookie, with *or without* a leading dot, makes it apply to all subdomains. Only not specifying it at all makes it apply only to the exact domain of the request. (Except for Internet Explorer, in which it always applies to subdomains anyway.) I long assumed that specifying the exact current domain would limit the cookie to that domain, but that's actually backwards, it un-limits it. (I can't take credit for any of this code though, it's pretty much all Jehiah ;)

reedloden:
@jehiah, anything else you'd like to see in/for this PR in order to get it landed?

reedloden:
Hey @jehiah -- anything I can do to help get this landed? This is blocking the use of cookie prefixes with `oauth2_proxy`, so would love to see it fixed. :)

PR #377, opened 2017-04-20
Refactor proxy director functions

travisofthenorth:
I noticed that these two proxy director functions do the same thing in terms of manipulating the request URL. Rather than have the code in both director functions, I moved it to the `Proxy` function where it will always run. Now the code only uses the custom director to pass the `Host`.

I wanted the ability to update this director code a bit, i.e. allow a couple more options in terms of manipulating the request. Maintaining both functions makes this a bit more cumbersome, hence the refactor.


travisofthenorth:
Whoops, need to fix the test failures.

PR #381, opened 2017-04-24
dist.sh: use go build option to strip binaries

ploxiln:
30% release binary size reduction - from 10487848 bytes to 7041472 bytes for the linux x86_64 build.

I also bumped the version for dev, I can rebase and omit that commit if a different version bump is done in master.


ploxiln:
Nope, this is all I had in mind. (Any requests?)

PR #382, opened 2017-04-24
README: nginx auth_request example updates

ploxiln:
Simplified, then added cookie handling. See commit messages for some rationale.

Any opinions @semenko @ashkulz ?


jehiah:
Going to consider conversation in #391 as validation here =)

PR #387, opened 2017-05-05
On failure, avoid immediately panic via nil pointer deref

virusdave:
Bad command line config currently panics.  This clearly isn't intended, so short-circuit the panic and let the existing configuration error code handle the error.


jehiah:
thanks @virusdave Can you add a test case that covers this situation too?

virusdave:
Sure.  Give me a little bit to get to it though

talam:
@virusdave going to close this out and push this through in https://github.com/bitly/oauth2_proxy/pull/502

PR #388, opened 2017-05-09
JSON userinfo endpoint

travisofthenorth:
This PR adds a JSON `/proxyPrefix/userinfo` endpoint so that JS can fetch the user data after authentication.

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	b64 "encoding/base64"
+	"encoding/json"
 	"errors"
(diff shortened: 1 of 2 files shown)


travisofthenorth:
Getting a failure on `go test -race` in go 1.7
```
2017/05/09 19:52:04 validator.go:42: failed opening authenticated-emails-file="/tmp/test_auth_emails_806575822", open /tmp/test_auth_emails_806575822: no such file or directory
```

Seems unrelated, so if anyone has any ideas on what might be causing it please advise.

PR #165, opened 2015-11-09
providers: add OpenID Connect provider

philips:
See the README for usage with Dex or any other OIDC provider.

This has only been tested against CoreOS Dex but should work against
other providers like Google, Facebook, etc as well.

To try it out see the improved Dex docs here: https://github.com/coreos/dex/pull/176

PR #389, opened 2017-05-09
*: add an OpenID Connect provider

### LINKED ISSUE #332: Consider adding dex as a provider
CoreOS have written the most excellent [dex](https://github.com/coreos/dex) web service for allowing oauth2 tokens to come from things like a corporate Active Directory or IPA LDAP server. Consider supporting dex as a provider.

ericchiang:
Takes over from https://github.com/bitly/oauth2_proxy/pull/165

Closes #332

cc @philips @gtaylor

See the README for usage with Dex or any other OIDC provider.

To test run a backend:

    python3 -m http.server

Run dex and modify the example config with the oauth2 callback.

    go get github.com/coreos/dex/cmd/dex
    cd $GOPATH/src/github.com/coreos/dex
    sed -i.bak \
      's|http://127.0.0.1:5555/callback|http://127.0.0.1:5555/oauth2/callback|g' \
       examples/config-dev.yaml
    make
    ./bin/dex serve examples/config-dev.yaml

Then run the oauth2_proxy:

    oauth2_proxy \
      --oidc-issuer-url http://127.0.0.1:5556/dex \
      --upstream http://localhost:8000 \
      --client-id example-app \
      --client-secret ZXhhbXBsZS1hcHAtc2VjcmV0 \
      --cookie-secret foo \
      --email-domain '*' \
      --http-address http://127.0.0.1:5555 \
      --redirect-url http://127.0.0.1:5555/oauth2/callback \
      --cookie-secure=false

Login with the username/password `admin@example.com:password`

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,22 @@ For adding an application to the Microsoft Azure AD follow [these steps to add a
 
 Take note of your `TenantId` if applicable for your situation. The `TenantId` can be used to override the default `common` authorization server with a tenant specific server.
 
+### OpenID Connect Provider
+
+OpenID Connect is a spec for OAUTH 2.0 + identity that is implemented by many major providers and several open source projects. This provider was originally built against CoreOS Dex and we will use it as an example.
+
+1. Launch a Dex instance using the [getting started guide](https://github.com/coreos/dex/blob/master/Documentation/getting-started.md).
+2. Setup oauth2_proxy with the correct provider and using the default ports and callbacks.
+3. Login with the fixture use in the dex guide and run the oauth2_proxy with the following args:
+
+    -provider oidc
+    -client-id oauth2_proxy
+    -client-secret proxy
+    -redirect-url http://127.0.0.1:4180/oauth2/callback
+    -oidc-issuer-url http://127.0.0.1:5556
+    -cookie-secure=false
+    -email-domain example.com
+
 ## Email Authentication
 
 To authorize by email domain use `--email-domain=yourcompany.com`. To authorize individual email addresses use `--authenticated-emails-file=/path/to/file` with one email per line. To authorize all email addresses use `--email-domain=*`.
--- a/options.go
+++ b/options.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"crypto"
 	"crypto/tls"
 	"encoding/base64"
@@ -14,6 +15,7 @@ import (
 
 	"github.com/18F/hmacauth"
 	"github.com/bitly/oauth2_proxy/providers"
+	oidc "github.com/coreos/go-oidc"
 )
 
 // Configuration Options that can be set by Command Line Flag, or Config File
@@ -63,6 +65,7 @@ type Options struct {
 	// These options allow for other providers besides Google, with
 	// potential overrides.
 	Provider          string `flag:"provider" cfg:"provider"`
+	OIDCIssuerURL     string `flag:"oidc-issuer-url" cfg:"oidc_issuer_url"`
 	LoginURL          string `flag:"login-url" cfg:"login_url"`
 	RedeemURL         string `flag:"redeem-url" cfg:"redeem_url"`
 	ProfileURL        string `flag:"profile-url" cfg:"profile_url"`
@@ -81,6 +84,7 @@ type Options struct {
 	CompiledRegex []*regexp.Regexp
 	provider      providers.Provider
 	signatureData *SignatureData
+	oidcVerifier  *oidc.IDTokenVerifier
 }
 
 type SignatureData struct {
@@ -120,6 +124,14 @@ func parseURL(to_parse string, urltype string, msgs []string) (*url.URL, []strin
 }
 
 func (o *Options) Validate() error {
+	if o.SSLInsecureSkipVerify {
+		// TODO: Accept a certificate bundle.
+		insecureTransport := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		http.DefaultClient = &http.Client{Transport: insecureTransport}
+	}
+
 	msgs := make([]string, 0)
 	if len(o.Upstreams) < 1 {
 		msgs = append(msgs, "missing setting: upstream")
@@ -135,6 +147,22 @@ func (o *Options) Validate() error {
 	}
 	if o.AuthenticatedEmailsFile == "" && len(o.EmailDomains) == 0 && o.HtpasswdFile == "" {
 		msgs = append(msgs, "missing setting for email validation: email-domain or authenticated-emails-file required.\n      use email-domain=* to authorize all email addresses")
+	}
+
+	if o.OIDCIssuerURL != "" {
+		// Configure discoverable provider data.
+		provider, err := oidc.NewProvider(context.Background(), o.OIDCIssuerURL)
+		if err != nil {
+			return err
+		}
+		o.oidcVerifier = provider.Verifier(&oidc.Config{
+			ClientID: o.ClientID,
+		})
+		o.LoginURL = provider.Endpoint().AuthURL
+		o.RedeemURL = provider.Endpoint().TokenURL
+		if o.Scope == "" {
+			o.Scope = "openid email profile"
+		}
 	}
 
 	o.redirectURL, msgs = parseURL(o.RedirectURL, "redirect", msgs)
@@ -210,13 +238,6 @@ func (o *Options) Validate() error {
 	msgs = parseSignatureKey(o, msgs)
 	msgs = validateCookieName(o, msgs)
 
-	if o.SSLInsecureSkipVerify {
-		insecureTransport := &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-		}
-		http.DefaultClient = &http.Client{Transport: insecureTransport}
-	}
-
 	if len(msgs) != 0 {
 		return fmt.Errorf("Invalid configuration:\n  %s",
 			strings.Join(msgs, "\n  "))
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -139,6 +139,22 @@ For adding an application to the Microsoft Azure AD follow [these steps to add a
...
 
+### OpenID Connect Provider
+
+OpenID Connect is a spec for OAUTH 2.0 + identity that is implemented by many major providers and several open source projects. This provider was originally built against CoreOS Dex and we will use it as an example.
+
+1. Launch a Dex instance using the [getting started guide](https://github.com/coreos/dex/blob/master/Documentation/getting-started.md).
+2. Setup oauth2_proxy with the correct provider and using the default ports and callbacks.
+3. Login with the fixture use in the dex guide.
+
+    -provider oidc
```
gtaylor: Can these be more explicitly marked as `oauth2_proxy` args? It makes sense if you've played with oauth2_proxy before, but may help those who are early in their evaluation/have yet to use it.

ericchiang: README section was just copied over from the previous PR. Let me revamp it a bit. Thanks.

options.go
```diff
@@ -137,6 +141,22 @@ func (o *Options) Validate() error {
 		msgs = append(msgs, "missing setting for email validation: email-domain or authenticated-emails-file required.\n      use email-domain=* to authorize all email addresses")
 	}
 
+	if o.OIDCIssuerURL != "" {
```
ericchiang: Woops, needs to come after overriding http.DefaultClient.

ericchiang:
@jehiah any chance someone at bitly has bandwidth to chime in on this PR?

mattenklicker:
@ericchiang I tried to use the oidc provider with dex, but without success.
After successful login in dex, oauth2-proxy tries to redeem the token directly from dex and this fails (it gets an error from dex):
```
2017/05/18 10:50:19 oauthproxy.go:527: 10.244.13.69:45572 ("xxxxx") error redeeming code token exchange: oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_request","error_description":"redirect_uri did not match URI from initial request."}
```
I patched dex to see what is in `redirect_uri` (btw: debug logging isn't working in dex) and it's empty.
After looking with tcpdump, it seems it's unset:
```
...9RJ.*POST /token HTTP/1.1
Host: xxxxxx
X-Real-IP: xxxx
Connection: close
X-Forwarded-For: xxxx, xxxxx
X-Forwarded-Host: xxxx
X-Forwarded-Port: 443
X-Forwarded-Proto: https
X-Original-URI: /token
X-Scheme: https
Content-Length: 61
X-Amzn-Trace-Id: Root=xxxxxx
content-type: application/x-www-form-urlencoded
authorization: Basic xxxxxx
accept-encoding: gzip
user-agent: Go-http-client/2.0

code=xxxxxx&grant_type=authorization_code
```

Do you have any idea why?

BTW: -redeem-url seems to be ingnored

ericchiang:
@mattenklicker --redeem-url is implied from discovery metadata. There aren't spec compliant reasons to override it.

The error your seeing implies that the URL that the bitly proxy is sending as a redirect URI isn't registered with dex. This is the same issue you'd have with google, github, etc. if you didn't register the correct redirect. What flags are you running with and what does your `staticClients` dex config look like?

bison:
I don't actually see the redirect URL being set in the `oauth2.Config` inside the `Redeem()` method. Am I just missing something?

ericchiang:
@bison the oauth2.Config is just used for token exchange. The Redeem URL reuses the proxy's built in OAuth2 redirects and is set here: https://github.com/bitly/oauth2_proxy/pull/389/files#diff-4aae454123f50b906f9aaef44f10011eR154

bison:
@ericchiang right, but the the `redirect_uri` is sent to the provider during the exchange. It needs to be set there.

mattenklicker:
@ericchiang the auth chain is: app -> oauth2-proxy -> dex -> ldap

app is a kubernetes ingress config
app ingress:
```yaml
metadata:
  annotations:
    ingress.kubernetes.io/auth-signin: https://$host/oauth2/sign_in
    ingress.kubernetes.io/auth-url: https://$host/oauth2/auth
spec:
  rules:
  - host: kubernetes-dashboard....
    http:
      paths:
      - backend:
          serviceName: kubernetes-dashboard
          servicePort: 80
        path: /
```
and auth redirect ingress:
```yaml
spec:
  rules:
  - host: kubernetes-dashboard....
    http:
      paths:
      - backend:
          serviceName: oauth2-proxy
          servicePort: 80
        path: /oauth2
```

oauth2-proxy config:
```yaml
       args:
        - --provider=oidc
        - --redirect-url=https://oauth2-proxy..../oauth2/callback
        - --email-domain=*
        - --upstream=file:///dev/null
        - --http-address=0.0.0.0:4180
        - --cookie-secure=false
        - --oidc-issuer-url=https://dex....
        env:
        - name: OAUTH2_PROXY_CLIENT_ID
          value: oauth2_proxy
        - name: OAUTH2_PROXY_CLIENT_SECRET
          value: proxy
        - name: OAUTH2_PROXY_COOKIE_SECRET
          value: secret
```

dex static clients config:
```yaml
staticClients:
- id: oauth2_proxy
  secret: proxy
  name: 'Oauth2-Proxy'
  redirectURIs:
  - 'https://oauth2-proxy..../oauth2/callback'
```

Note: The app-callback-URL itself is nowhere registered. But I don't think that's the problem here. If it should be registered, than it should be with oauth2-proxy (?). I tried to add it to `redirectURIs` without success.

Another question: `--skip-provider-button` let me guess, that this would skip the oauth2-proxy login page and would redirect directly to dex's login page. But it has no effect.

benley:
I'm stuck on the same invalid_request error that @mattenklicker and others have mentioned.  Does anyone know of a way to get this working?

bison:
I still think that's because the `redirect_uri` is not being sent when requesting an access token. It needs to be set on the `oauth2.Config`. Otherwise [this](https://github.com/coreos/dex/blob/master/server/handlers.go#L637-L640) happens.

benley:
Yep, sure enough.  This is a one-line fix, and now it works:

```patch
diff --git a/providers/oidc.go b/providers/oidc.go
index 580c1e3..98d9220 100644
--- a/providers/oidc.go
+++ b/providers/oidc.go
@@ -23,6 +23,7 @@ func NewOIDCProvider(p *ProviderData) *OIDCProvider {
 func (p *OIDCProvider) Redeem(redirectURL, code string) (s *SessionState, err error) {
        ctx := context.Background()
        c := oauth2.Config{
+               RedirectURL:  redirectURL,
                ClientID:     p.ClientID,
                ClientSecret: p.ClientSecret,
                Endpoint: oauth2.Endpoint{
```

mattenklicker:
Nice. Works fine.
For documentation: You can omit `--redirect-url=....`, so you can use one proxy for all your apps.

emmanuel:
As an OIDC (and OAuth) n00b, I'm a bit disoriented about where this PR stands. Can someone help me with the below questions?

1. Is the PR held up from being merged by the aforementioned one line change? 
2. Does it work without that change so long as `--redirect-url` is omitted?
3. What about the additional changes over in https://github.com/postmates/oauth2_proxy/commits/oidc-provider? (packaging, but also passing raw token in authorization header... important?)

benley:
@emmanuel I added the Authorization header because it makes oauth2_proxy work properly with the Kubernetes dashboard app, which uses the bearer token to act on the behalf of the authenticated user.  It's probably not needed for most other apps.

emmanuel:
@benley thanks for the explanation. that sounds like a good use-case (i.e., I'm off to deploy the proxy in front of kube-dashboard 😁 ).

providers/oidc.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (p *OIDCProvider) Redeem(redirectURL, code string) (s *SessionState, err error) {
+	ctx := context.Background()
+	c := oauth2.Config{
+		ClientID:     p.ClientID,
+		ClientSecret: p.ClientSecret,
+		Endpoint: oauth2.Endpoint{
+			TokenURL: p.RedeemURL.String(),
+		},
+	}
```
travisofthenorth: Doesn't this config need the `RedirectURL`? I was getting a 401 without it.

kongslund: I encountered the same as @travisofthenorth.
@ericchiang ?

ericchiang:
Comments addressed. Please let me know if there's anything else blocking this PR.

travisofthenorth:
My feedback has been addressed 👍

philips:
@travisofthenorth great! Who can click merge?

benley:
:1st_place_medal:

philips:
🤣

kfox1111:
I've almost got this working in front of kube-dashboard. I get authenticated through the proxy but the handoff of the token doesn't seem to make it to kube-dashboard... I see in the kube-dashboard 1.7 docs it mentions bearer tokens. I don't really see how they are provided in this ps though. Am I missing a flag? Any other ideas?

benley:
Are you using a build that's patched to pass the user's raw token along in a Authorization header? Here's what finally got that part working for me: https://github.com/postmates/oauth2_proxy/commit/5a27234e167a81f4bccd668a7171b797289d3db0

kfox1111:
I checked out trunk from earlier today. Is there still patches needing to be merged?

kfox1111:
yeah.... looks like that other patch is not here yet... Thanks for the pointer to it. Any plans for a pr here?

kfox1111:
I could see other things wanting to use bearer tokens too? Seems like a very minor thing to fork a project over. If upstream will take it, probably worth a shot?

kfox1111:
tried to cherry pick the one patch. it fails due to:
./oauthproxy.go:690:61: session.RawIDToken undefined (type *providers.SessionState has no field or method RawIDToken)

Looks like the two code bases are farther apart then I had hoped. :(

kfox1111:
I think I figured it out. 'go get' was fetching a new checkout of the bitly/oauth2_proxy rather then reusing the existing, patched checkout. After patching the downloaded source, the patch seems to have worked as is.

andreychernih:
Any chance we can get this patch in the upstream?

emmanuel:
I would also be interested in this patch being incorporated in the official project, for the same purpose[1]. Basically, "would it hurt to include it upstream"?

[1] for use with the Kubernetes dashboard, which requires the actual auth JWT to enable the dashboard to impersonate/operate-on-behalf-of the logged in user

Drupi:
Hello ! Can some1 share me a container with this PR merged

PR #392, opened 2017-05-18
[github provider] use Authorization header, not access_token query parameter

arnottcr:
The use of the `access_token` GET query parameters when making api calls to github increases the possibility of accidentally logging the full text access token; since github supports passing these tokens via header it makes sense to do that, prevents the rick of disclosure and removes the need for a `tokenStrip()` method.

### DIFF
--- a/providers/github.go
+++ b/providers/github.go
@@ -62,8 +62,7 @@ func (p *GitHubProvider) hasOrg(accessToken string) (bool, error) {
 	}
 
 	params := url.Values{
-		"access_token": {accessToken},
-		"limit":        {"100"},
+		"limit": {"100"},
 	}
 
 	endpoint := &url.URL{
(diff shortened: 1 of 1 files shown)


arnottcr:
this might also be a decent thing to do with other oauth providers; GET parameters are just not a great place to put credentials

ploxiln:
Looks good (great!) to me. I'll test it myself in the next few days.

arnottcr:
I can mock up something for other providers (I am thinking gitlab atm), as this is a thing that seems portable; let me know your thoughts.

ploxiln:
I have now tested this myself with a real GitHub oauth2 app, and it works for me.

I think this should be done for other providers in separate pull requests, so they can be tested/validated by users of those providers.

arnottcr:
cool, that makes sense; when I get a second, I will see what other providers support header auth and make seperate PRs for them.

PR #395, opened 2017-05-22
Fix handling of redirect URI with --skip-provider-button

### LINKED ISSUE #327: skip-provider-button seems to force final redirection url to be /
When I use -skip-provider-button it always gets redirected after login to /
When I don't it gets redirected after login to the url beforehand.

e.g. I point my browser at https://my-site/some/sub/path
when I have -skip-provider-button, it will end up pointing at https://my-site/
when I don't it will end up pointing at https://my-site/some/sub/path

I have the oauth2_proxy behind an apache server.
I use the following command (of course I've censored/changed some):
`oauth2_proxy -provider google -http-address http://:4180 -cookie-expire 1h  -client-id XXXXXX -client-secret XXXXXX -cookie-secret XXXXX -email-domain my-site -login-url https://accounts.google.com/o/oauth2/auth?access_type=offline&hd=my-site -upstream http://private-server:80/`

Note: I'm using a proper domain name where you see 'my-site' above.

Note: I supplied my own login-url, so I could add the 'hd' parameter to have google only prompt with accounts from the given organization/domain.

bastianeicher:
Fixes #327

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -410,6 +410,11 @@ func (p *OAuthProxy) ManualSignIn(rw http.ResponseWriter, req *http.Request) (st
 }
 
 func (p *OAuthProxy) GetRedirect(req *http.Request) (redirect string, err error) {
+	if p.SkipProviderButton {
+		redirect = req.RequestURI
+		return
+	}
+
 	err = req.ParseForm()
 	if err != nil {
 		return


bastianeicher:
Travis build failed because of `no space left on device`. Does not seem to be a problem with the commit itself.

ploxiln:
You might need to add a bit more of the logic which the sign_in page uses for this: https://github.com/bitly/oauth2_proxy/blob/v2.2/oauthproxy.go#L367-L373

rmoriz:
@ploxiln 
So this PR does not yet support  the HTTP header "X-Auth-Request-Redirect" and therefore isn't merged, right?

adrianchifor:
Any progress on this?

ploxiln:
this repo is abandoned, there are a couple successors, see https://github.com/bitly/oauth2_proxy/issues/628#issuecomment-417121636

PR #405, opened 2017-06-21
Update Google Auth Provider instructions

bspaans:
"Enable and manage APIs" is now called "API Manager"


jehiah:
Thanks @bspaans !

Issue #284, opened 2016-07-25
Removing login prompt


Totolin:
Is there any possibility to remove the webpage that prompts you to log in using the provider set in the configuration? It really has no purpose if you have only one provider, and it should redirect straight to the oauth provider's page.

Is there an option to do this, or have I missed it in the documentation?

ploxiln:
`--skip-provider-button`
https://github.com/bitly/oauth2_proxy/blob/2fa0a35c7f89706ddf9dff3e6956a1af8b937066/main.go#L39

reist:
In case of using Nginx's auth_request, should the /oauth2/sign_in be switched to /oauth2/start to get the same effect? Based on a couple of checks, it seems to be working as long as the _oauth2_proxy cookie didn't timeout.

Alives:
This option isn't working for me.  I only have Google configured as a provider, and I still see the "Sign in with a Google Account" button page if I don't have the oauth cookie set properly.

PR #406, opened 2017-06-21
providers: add okta

abraithwaite:


oauthproxy.go
```diff
@@ -482,7 +482,11 @@ func (p *OAuthProxy) SignIn(rw http.ResponseWriter, req *http.Request) {
 		p.SaveSession(rw, req, session)
 		http.Redirect(rw, req, redirect, 302)
 	} else {
-		p.SignInPage(rw, req, 200)
+		if p.SkipProviderButton {
+			p.OAuthStart(rw, req)
+		} else {
+			p.SignInPage(rw, req, 200)
+		}
```
abraithwaite: I think this is what #284 is referring to.

abraithwaite:
Could still use some tests.  Lemme know if you have any specific rules about contributing and I'm happy to comply.

oauthproxy.go
ploxiln: It would be better for this to be in a separate PR, not mixed in with this one.

Fodoj:
This is very much needed!

asms:
@abraithwaite Nice!

statik:
See issue #428

asms:
@abraithwaite Validation does not work. Specifically request to `/oauth2/v1/userinfo` returns a 400 Bad Request. The reason is that the access_token is passed as a GET argument rather than using the `Authorization: Bearer ...` header. At a glance, it seems like `ValidateSessionState ` needs to be implemented (see https://github.com/bitly/oauth2_proxy/blob/a9c55bd6d1caeaa7ba20236b2e85129c8219b865/providers/facebook.go#L78 for an example).

abraithwaite:
Thanks.  I'll try and look at it when I have time, otherwise if someone else beats me to it please close this PR!

rothgar:
Commenting here to show support for this and follow updates

rothgar:
I got this built but having trouble setting up the okta side of the application for testing. I think I'm not setting up my login redirect URI properly but if someone has an example it'd be helpful.
My requests are redirecting properly but I'm getting this on Okta
image

I see the request redirect to `GET - "/oauth2/start?rd=%2F" HTTP/1.0` but I still receive a 400 from okta.

reedloden:
Since I can't edit this PR directly, opened #528.

abraithwaite:
Neat!  Been waiting to close this.  Thanks!

PR #407, opened 2017-06-21
nginx auth_request: fix -skip-provider-button

### LINKED ISSUE #284: Removing login prompt

abraithwaite:
fixes #284

### DIFF
--- a/oauthproxy_test.go
+++ b/oauthproxy_test.go
@@ -3,9 +3,6 @@ package main
 import (
 	"crypto"
 	"encoding/base64"
-	"github.com/18F/hmacauth"
-	"github.com/bitly/oauth2_proxy/providers"
-	"github.com/bmizerany/assert"
 	"io"
 	"io/ioutil"
 	"log"
@@ -17,6 +14,10 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/18F/hmacauth"
+	"github.com/bitly/oauth2_proxy/providers"
+	"github.com/bmizerany/assert"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


oauthproxy_test.go
```diff
@@ -409,20 +414,48 @@ func TestSignInPageIncludesTargetRedirect(t *testing.T) {
...
+	sip_test := NewSignInPageTest(true)
+	const endpoint = "/some/random/endpoint"
+
+	code, body := sip_test.GetEndpoint(endpoint)
+	assert.Equal(t, 302, code)
+
+	match := sip_test.sign_in_provider_regexp.FindStringSubmatch(body)
+	if match == nil {
+		t.Fatal("Did not find pattern in body: " +
+			signInRedirectPattern + "\nBody:\n" + body)
```
abraithwaite: Fixed the wrong one.

ploxiln:
Just to clarify: If in proxy mode, the `/sign_in` route should never be hit, because proxy does this:

```go
func (p *OAuthProxy) Proxy(rw http.ResponseWriter, req *http.Request) {
	status := p.Authenticate(rw, req)
	if status == http.StatusInternalServerError {
		p.ErrorPage(rw, http.StatusInternalServerError,
			"Internal Error", "Internal Error")
	} else if status == http.StatusForbidden {
		if p.SkipProviderButton {
			p.OAuthStart(rw, req)
		} else {
			p.SignInPage(rw, req, http.StatusForbidden)
		}
	} else {
		p.serveMux.ServeHTTP(rw, req)
	}
}
```

So the case that this fixes is if "nginx auth_request" mode is being used, and the user is setting the nginx `error_page` to `/oauth2/sign_in`. They can already get the behavior they want by setting the nginx `error_page` to `/oauth2/start` instead.

abraithwaite:
Yep, doesn't really make a huge difference to me.  I'll try it by setting the `error_page` as it seems more clean.

Just thought it might make sense here just for consistency with the other path and for newbies like me who get confused when that setting doesn't work with the example nginx conf. :-)

ploxiln:
yes probably still a worthwhile change

Issue #411, opened 2017-06-29
auth_request endpoint should always allow sign in and callback page


dereulenspiegel:
I am currently trying to use oauth2_proxy (great software btw.) in combination with an nginx based ingress in kubernetes. I use oauth2_proxy as the server for auth_request based authentication. Unfortunately the ingress resource does not allow me to exclude certain paths from authentication. So all calls to i.e. /oauth2/sign_in are authenticated by the oauth2_proxy which, of course, denies the request, since I havent't signed in yet.
Does make it sense to let oauth2_proxy allow all requests to urls like <prefix>/sign_in and <prefix>/callback? Or will this break any existing use cases or security?

PR #412, opened 2017-07-03
Inspect X-Original-URI header and allow internal URLs

dereulenspiegel:
This PR helps with #411. If you send your original request uri via the header `X-Original-URI` it will be taken into consideration when authenticating the request. This makes it easier to use oauth2_proxy as auth_request service with nginx. For example the nginx based ingress for kubernetes doesn't allow you to enforce authentication only for selected locations, so oauth2_proxy needs to allow the oauth2 callback urls etc.
Here is an example ingress config for use with oauth2_proxy
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: internal-ingress
  annotations:
    ingress.kubernetes.io/auth-url: "http://oauth2-proxy.default.svc.cluster.local:8080/oauth2/auth"
    ingress.kubernetes.io/auth-method: "GET"
    ingress.kubernetes.io/auth-response-headers: X-Auth-Request-User, X-Auth-Request-Email
      add_header Set-Cookie $auth_cookie;
      auth_request_set $auth_cookie $upstream_http_set_cookie;
      auth_request_set $user   $upstream_http_x_auth_request_user;
      auth_request_set $email  $upstream_http_x_auth_request_email;
      proxy_set_header X-User  $user;
      proxy_set_header X-Email $email;
      error_page 401 /oauth2/sign_in;

    kubernetes.io/ingress.class: "nginx"
    ingress.kubernetes.io/force-ssl-redirect: "true"
    ingress.kubernetes.io/ssl-redirect: "true"
spec:
...
```

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -594,6 +594,19 @@ func (p *OAuthProxy) Proxy(rw http.ResponseWriter, req *http.Request) {
 }
 
 func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) int {
+	// if any of the internal urls necessary for the oauth2 flow is called, allow the request
+	// This is necessary to use oauth2_proxy as auth_request server for nginx based k8s ingresses
+	originalRequestUri := req.Header.Get("X-Original-URI")
+	for _, allowedUrl := range []string{
+		p.SignInPath,
+		p.SignOutPath,
+		p.OAuthCallbackPath,
+		p.OAuthStartPath,
+	} {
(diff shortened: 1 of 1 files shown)


ploxiln:
An alternative, for your special case, would be to use `oauth2_proxy` in the basic "proxy mode" (instead of "auth_request" mode) in _front_ of the kubernetes ingress.

xeor:
I am using oauth2 with kubernetes as well, this is much needed!

@dereulenspiegel as a side question:
Would your example populate the x-user header the kubernetes apps sees?
Example with the mail-address of the user if using the google oauth provider?

PR #414, opened 2017-07-04
Iterate through pages returned by List Your Organizations endpoint

relaxdiego:
PREAMBLE: This PR leaves much to be desired at the time of submission.
I'm submitting it anyway to get early feedback. I'm also still ramping up
on my golang-fu so please excuse the dust.

=====

For some GHE instances where a user can have more than 100
organizations, traversing the other pages is important otherwise
oauth2_proxy will consider the user unauthorized. This change traverses
the list returned by the API to avoid that.

### DIFF
--- a/providers/github.go
+++ b/providers/github.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


providers/github.go
```diff
@@ -61,36 +62,45 @@ func (p *GitHubProvider) hasOrg(accessToken string) (bool, error) {
 		Login string `json:"login"`
 	}
 
-	params := url.Values{
-		"limit": {"100"},
+	var orgs_page []struct {
```
hlhendy: Comment on naming, in go we typically use camel case/mixed caps rather than underscores - so I'd update this to `orgsPage` and page_number to `pageNum`.

https://golang.org/doc/effective_go.html#mixed-caps

talam:
@hlhendy update this PR based on the last set of comments and made some slight updates to the logic. Also included some tests. RFR.

PR #425, opened 2017-07-20
Update cookie generation to match base64 encoding

bluecmd:
Current code is using URLEncoding but example was using the
standard RFC 4648 encoding. Switch to using the URL
encoding in the example as well.


ploxiln:
hmm sure is: https://github.com/bitly/oauth2_proxy/blob/v2.2/options.go#L304
thanks!

PR #426, opened 2017-07-20
Remove check for >0 upstreams

bluecmd:
When used solely for auth_request there is no upstream.
Instead of forcing users to set a dummy upstream, remove
the check.


ploxiln:
Makes sense to me - could log a warning though.

Looks like you need to update `TestNewOptions`.

bluecmd:
@ploxiln Right you are. Amended the previous commit, please have a look!

ploxiln:
👍

Issue #399, opened 2017-06-01
Support for a whitelist of `--redirect-domains`


colemickens:
Per my comments [here](https://github.com/bitly/oauth2_proxy/pull/360), I'd like to add support for `--redirect-domains` to take a comma-delimited list of whitelisted redirect domains.

Example scenario:

* https://secret-static-content.mydomain.tld -> `httpd` fronting a static directory
* https://more-secret-content.mydomain.tld -> `httpd` fronting a different static directory
* https://auth.mydomain.tld/oauth2 -> oauth2_proxy instsance

Currently, I can't seem to get things to work properly in terms of handling the redirect after the callback succeeds.

It seems like it's due to the filtering that happens here:

* https://github.com/bitly/oauth2_proxy/blob/6d6cb7e1f83f0b9bb96bef14907bfbc5d128cf02/oauthproxy.go#L419

The same logic seems duplicated here:

* https://github.com/bitly/oauth2_proxy/blob/6d6cb7e1f83f0b9bb96bef14907bfbc5d128cf02/oauthproxy.go#L551

I propose a PR such that:

* there is a new `--redirect-domains` flag that takes a list of domains. example: (`--redirect-domains=mickens.io,mickens.us`).

* the redirect filter is extended to allow any URLs that have a Host that ends with a value in the `--redirect-domain` list.

Does this sound reasonable?

colemickens:
Actually, I'm not sure I understand why the redirect locations needs to be limited... is that a security concern as well? Thanks!

jehiah:
@colemickens Thanks for this PR; i'll hopefully get to look through it soon.

There is a security aspect to restricting redirect domains. This is to protect against an attacker crafting a URL with a malicious redirect parameter that will send you through the oauth2_proxy auth flow as you would expect, but then redirect to a malicious site after authentication. A user should only end on an authorized URL after authentication. This eliminates many avenues of phishing and other social engineering.

ploxiln:
also discussed in #378

colemickens:
Not a PR yet, just curious if it would be desired/accepted before I write it. I've confirmed that just removing those filtering lines mentioned unblocks my scenario, but I don't want to continue to use that if it's insecure and would prefer to just use upstream instead of my fork anyway.

[...]

PR #427, opened 2017-07-22
Allow redirect URL to be passed in the query string.

johnbelamaric:
When using the Kubernetes nginx ingress controller, the sign_in page is reached via a 302 redirect, which means that we cannot use the `X-Auth-Request-Redirect` header. Instead, we need to be able to include the URL in the query string.

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -368,6 +368,9 @@ func (p *OAuthProxy) SignInPage(rw http.ResponseWriter, req *http.Request, code 
 	if req.Header.Get("X-Auth-Request-Redirect") != "" {
 		redirect_url = req.Header.Get("X-Auth-Request-Redirect")
 	}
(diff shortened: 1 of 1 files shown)


aledbf:
Any update on this?

ploxiln:
* you could just redirect to `.../oauth2/start?rd=...` instead
  * if you really want this here, you should use `GetRedirect()` here (it does a bit more to prevent "open-redirects")

aledbf:
@ploxiln is possible a redirect to a different host?

ploxiln:
see #399

aledbf:
@ploxiln ok, so the recommendation is to not allow "external" redirects or adding the whitelist domain flag is ok?

ploxiln:
A domain whitelist feature would be as good as the current status-quo from a security/phishing perspective. It's just arbitrary (completely unrestricted) redirect domain which is problematic.

PR #436, opened 2017-08-08
Bcrypt Support

tw:


reedloden:
Any reason why this got closed? Seems like a legit PR. :)

PR #446, opened 2017-09-08
Options: Update github-org arg to support a list of orgs

schottkey7:


oauthproxy.go
```diff
@@ -684,7 +684,12 @@ func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) int
 		if session.Email != "" {
 			rw.Header().Set("X-Auth-Request-Email", session.Email)
 		}
+
+		if p.PassAccessToken && session.AccessToken != "" {
+			rw.Header().Set("X-Auth-Request-Access-Token", session.AccessToken)
+		}
 	}
```
schottkey7: Had to do this because setting the command line arg --pass-access-token=true and setting 

```
auth_request_set $access_token $upstream_http_x_forwarded_access_token;
```
and
```
proxy_set_header X-Forwarded-Access-Token $access_token;
```

in Nginx didn't set the X-Forwarded-Access-Token header.

PR #447, opened 2017-09-12
Use read_user as default scope for GitLab

Miouge1:
"read_user" works as a default scope for GitLab. The previous "api" scope gives full API access AFAIK?


jehiah:
thanks @Miouge1

Issue #450, opened 2017-09-14
Basic Auth to authenticate for token server?


deitch:
I have been using oauth2_proxy successfully against a variety of the supported providers, but also against a custom provider just by setting the right vars:

* `redirect_url`
* `login_url`
* `redeem_url`
* `scope`

And a few of the usual others.

The problem I have run into is that oauth2_proxy by default appears to try to authenticate by putting the `client_id` and `client_secret` into the query params rather than using HTTP Basic Auth. See e.g.[Google Redeem()](https://github.com/bitly/oauth2_proxy/blob/master/providers/google.go#L106) and [default Redeem()](https://github.com/bitly/oauth2_proxy/blob/master/providers/provider_default.go#L21), which look like:

```go
	params.Add("client_id", p.ClientID)
	params.Add("client_secret", p.ClientSecret)
``` 

Yet the [RFC specifies](https://tools.ietf.org/html/rfc6749#section-2.3.1) that an oauth2 token server MUST support http basic auth, and that it MAY also support it in params, but is not recommended:

```
The authorization server MUST support the HTTP Basic
   authentication scheme for authenticating clients that were issued a
   client password.

   For example (with extra line breaks for display purposes only):

     Authorization: Basic czZCaGRSa3F0Mzo3RmpmcDBaQnIxS3REUmJuZlZkbUl3

   Alternatively, the authorization server MAY support including the
   client credentials in the request-body using the following
   parameters:

   client_id
         REQUIRED.  The client identifier issued to the client during
         the registration process described by Section 2.2.

   client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.




Hardt                        Standards Track                   [Page 16]
 
RFC 6749                        OAuth 2.0                   October 2012


   Including the client credentials in the request-body using the two
   parameters is NOT RECOMMENDED and SHOULD be limited to clients unable
   to directly utilize the HTTP Basic authentication scheme (or other
   password-based HTTP authentication schemes).  The parameters can only
   be transmitted in the request-body and MUST NOT be included in the
   request URI.
```

Many implementations only support HTTP Basic.

1. Is this supported somewhere and I missed it?
2. If not, the change looks simple, would a PR be accepted?

jehiah:
@deitch this is similar to #392. I'd be happy to accept PR's to update any for providers that do support `HTTP Basic`.

deitch:
[...]

PR #451, opened 2017-09-17
Add support for Basic HTTP for Redeem() on a per-provider basis

deitch:
As discussed in https://github.com/bitly/oauth2_proxy/issues/450

This PR adds the ability for a provider to flag that it wants to pass `client_secret` using http basic auth, rather than in the `POST` body. According to the RFC, an implementation "MUST support the HTTP Basic authentication scheme for authenticating clients", and only "MAY support including the client credentials in the request-body". 

Once this is merged, we can test each provider and enable it by having it set `p.HTTPBasicAuth = true`. Right now, the default is `p.HTTPBasicAuth = false`. Eventually, this should be flipped (per the RFC).

Additional changes:

* adds unit tests for enabling/disabling `p.HTTPBasicAuth`
* extracts some common testing code from `providers/google_test.go` and `providers/provider_default_test.go` into `providers/common_test.go`
* adds an additional return value to `newRedeemServer()` of a `map` containing the body, headers and auth of the request, so it can be validated

### DIFF
--- a/options.go
+++ b/options.go
@@ -244,6 +244,7 @@ func (o *Options) Validate() error {
 
 func parseProviderInfo(o *Options, msgs []string) []string {
 	p := &providers.ProviderData{
+		HTTPBasicAuth:  false,
 		Scope:          o.Scope,
 		ClientID:       o.ClientID,
 		ClientSecret:   o.ClientSecret,
--- a/providers/common_test.go
+++ b/providers/common_test.go
@@ -1,0 +1,37 @@
+package providers
+
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+)
+
+type redeemResponse struct {
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+	ExpiresIn    int64  `json:"expires_in"`
+	IdToken      string `json:"id_token"`
+}
+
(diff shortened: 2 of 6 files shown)


deitch:
Any maintainers commenting?

deitch:
@jehiah still around? Can we get this through (or changes if necessary), and then I can start opening PRs for individual providers after testing? I opened the PR as requested.

options.go
```diff
@@ -244,6 +244,7 @@ func (o *Options) Validate() error {
 
 func parseProviderInfo(o *Options, msgs []string) []string {
 	p := &providers.ProviderData{
+		HTTPBasicAuth:  false,
```
ploxiln: this property name is a bit too generic - could be something like "RedeemWithBasicAuth"

deitch: Sure. Middle of working with a client on sleuthing a performance problem. Expect an update soon.

ploxiln:
Because the client id/secret is currently passed in a POST body, and that apparently works for all current providers, I think this change is not really worth doing right now. If a provider comes along that needs client id/secret in the redeem request Authorization header, they can override the default Redeem method. If a few of them do that we might consider changing the default for all of them.

Consider also the low rate of review by maintainers - there are other things that are more important for the limited time they give to this project.

deitch:
> Consider also the low rate of review by maintainers

Sigh, yeah.

> Because the client id/secret is currently passed in a POST body, and that apparently works for all current providers

Still, it would be nice to comply with the spec. And why wouldn't we want to support _any_ provider?

Issue #456, opened 2017-09-27
RD form value on login page


krogon-dp:
Correct rd form value when param not specified.
https://example.domain.com/oauth2/sign_in

    <form method="GET" action="/oauth2/start">
      <input type="hidden" name="rd" value="/">
	
      <button type="submit" class="btn">Sign in with a GitHub Account</button><br>
    </form>

Incorrect rd form value when param specified, making redirect to login_page again (loop)
https://example.domain.com/oauth2/sign_in?rd=/

    <form method="GET" action="/oauth2/start">
      <input type="hidden" name="rd" value="/oauth2/sign_in?rd=/">
	
      <button type="submit" class="btn">Sign in with a GitHub Account</button><br>
    </form>

Version: 2.2.1-alpha

```    spec:
      containers:
      - args:
        - --provider=github
        - --email-domain=*
        - --github-org=exampleOrganisation
        - --github-team=exampleGroup
        - --upstream=file:///dev/null
        - --cookie-domain=example.com
        - --cookie-name=_oauth2_proxy_test
        - --http-address=0.0.0.0:4180
        - --proxy-prefix=/oauth2
        image: docker.io/colemickens/oauth2_proxy:latest
       [...]

```

ploxiln:
Currently it's not supported to add `rd=...` param to the sign-in page url. You have two options:

  * if using original proxy mode, just trying to visit the intended final page while not authenticated will return the sign-in page with rd set correctly
  * if using "nginx auth_request" mode, set the `X-Auth-Request-Redirect` header, see https://github.com/bitly/oauth2_proxy#configuring-for-use-with-the-nginx-auth_request-directive

krogon-dp:
I see it now why it stopped working for me. nginx-intress-controller for kubernetes fixed the rd param handling in nginx-0.9.0-beta.12 version with this commit: https://github.com/kubernetes/ingress/commit/b2be9f0e040d4207aab80f088b9fdfe79954741e#diff-b7803798d356c6c17a90d93cc58bdbaaR588 

Correct me if I am wrong, but adding X-Auth-Request-Redirect header won't solve the problem unless rd param is still being set in `error_page 401` directive.
The only way is custom nginx template for ingress which reverts changes from above mentioned commit.

krogon-dp:
Reading code seems like `X-Auth-Request-Redirect` would work if set to static destination.

[...]

PR #457, opened 2017-09-27
Fixes: #456 #427

krogon-dp:


ploxiln:
This has the same problem as #427 has: it does not validate the redirect url is acceptable, and enables what is called an "open redirect".

(#427 is a better version of this because it properly parses the request for the rd parameter.)

Issue #458, opened 2017-09-29
Wrong version in Godeps file for cloud.google.com/go/compute/metadata


philippthun:
Due to an [issue in gpm](https://github.com/pote/gpm/issues/91), some packages are not checked out correctly. When manually checking out `cloud.google.com/go/compute/metadata` to version `v0.7.0` (as specified in the Godeps file), the following error is reported when running `go test`:

```
vendor/cloud.google.com/go/internal/retry.go:21:2: cannot find package "github.com/googleapis/gax-go" in any of:
	/Users/me/git/oauth2_proxy-boshrelease/src/github.com/bitly/oauth2_proxy/vendor/github.com/googleapis/gax-go (vendor tree)
	/usr/local/go/src/github.com/googleapis/gax-go (from $GOROOT)
	/Users/me/git/oauth2_proxy-boshrelease/src/github.com/googleapis/gax-go (from $GOPATH
```

talam:
resolved in https://github.com/bitly/oauth2_proxy/pull/459.

PR #459, opened 2017-09-29
Fix Godeps version for cloud.google.com/go/compute/metadata

philippthun:
When manually checking out `v0.8.0` instead of `v0.7.0` for `cloud.google.com/go/compute/metadata`, oauth2_proxy can be compiled and successfully tested.

This fixes issue #458.

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -7,7 +7,7 @@ gopkg.in/fsnotify.v1                     v1.2.0
 golang.org/x/oauth2                      7fdf09982454086d5570c7db3e11f360194830ca
 golang.org/x/net/context                 242b6b35177ec3909636b6cf6a47e8c2c6324b5d
 google.golang.org/api/admin/directory/v1 650535c7d6201e8304c92f38c922a9a3a36c6877
-cloud.google.com/go/compute/metadata     v0.7.0
+cloud.google.com/go/compute/metadata     v0.8.0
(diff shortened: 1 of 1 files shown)


philippthun:
Added implicit dependencies and sorted Godeps file.

ploxiln:
First of all, thanks for noticing this - gpm silently decides not to check out the specified tag for `cloud.google.com/go/compute/metadata`! The reason it didn't work is because gpm looks for a `.git` folder right in the imported gopath, not in a parent directory: https://github.com/pote/gpm/blob/master/bin/gpm#L77

This PR is a nice update overall but does not fix that key problem.

What's the best way to fix that problem? I'm not sure. Maybe if we specify just `cloud.google.com/go` in the `Godeps`?

philippthun:
It would fix the problem for me :smile: - I have a script that uses the `Godeps` file as input, issues `go get` commands and then does the `git checkout` (as all packages are using Git that just works fine). But of course, that's only a workaround. In my opinion this needs to be fixed in gpm. Thus I also created a PR for that: https://github.com/pote/gpm/pull/92

Specifying `cloud.google.com/go` would work, but does not seem right to me. I think the dependencies should be as specific as possible. But if the underlying issue in gpm does not get resolved, that will be the only possibility.

talam:
@philippthun thanks for attempting to resolve the issue but I am going to close this PR out as we transitioned over to using `dep` in https://github.com/bitly/oauth2_proxy/pull/491.

PR #460, opened 2017-09-30
Added Dockerfile

boivie:
It's written to generate a small image - around 16 MB.

Fixes #372

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM golang:1.9 AS builder
+
+WORKDIR /go/src/github.com/bitly/oauth2_proxy
+COPY . /go/src/github.com/bitly/oauth2_proxy/
+RUN go get -d -v
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo .
+
+FROM alpine:latest
+
+RUN apk --no-cache add ca-certificates
+WORKDIR /root/
+COPY --from=builder /go/src/github.com/bitly/oauth2_proxy/oauth2_proxy .
+
+ENTRYPOINT ["/root/oauth2_proxy"]


ploxiln:
Probably worth mentioning: I think this requires the very latest version of docker aka moby 17.05.0-ce

ploxiln:
hmm I take that back, it seems that's just when docker stopped tagging releases in the public github repo

boivie:
Yes, it's a new feature of Docker 17.05. That's why I recommend building it at Docker Cloud, which supports it.

yacut:
You cloud use scratch image to reduce image size to 30% :)
Stratch: 11.6MB
Alpine: 16.1MB

```sh
FROM golang:1.9 AS builder
WORKDIR /go/src/github.com/bitly/oauth2_proxy
COPY . .
RUN go get -d -v; \
    CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo . ; \
    curl -o ca-certificates.crt https://curl.haxx.se/ca/cacert.pem;

FROM scratch
COPY --from=builder /go/src/github.com/bitly/oauth2_proxy/oauth2_proxy /bin/oauth2_proxy
COPY --from=builder /go/src/github.com/bitly/oauth2_proxy/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt

ENTRYPOINT ["/bin/oauth2_proxy"]

```

jacobalberty:
Using boivie's dockerfile now works great hope to see official images soon.

Couldn't get the from scratch one working, kept giving errors relating to the cert chain. just a little tip though you don't need to curl the cert file as you can just use
```
ADD https://curl.haxx.se/ca/cacert.perm /etc/ssl/certs/ca-certificates.crt
```
to pull it instead.

jacobalberty:
Oh and one other suggested change. Add a `.dockerignore` file to the root that contains at least
```
Dockerfile
```
This will ensure editing your Dockerfile doesn't immediately invalidate virtually the entire build cache so if  you're testing things using a local built Dockerfile you can take advantage of the build cache to decide whether or not the whole thing needs to be rebuilt.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+COPY . /go/src/github.com/bitly/oauth2_proxy/
+RUN go get -d -v
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo .
+
+FROM alpine:latest
+
+RUN apk --no-cache add ca-certificates
+WORKDIR /root/
+COPY --from=builder /go/src/github.com/bitly/oauth2_proxy/oauth2_proxy .
+
```
rhuddleston: add "USER nobody" so we don't run as root

adrianchifor:
Any update?

PR #461, opened 2017-09-30
External redirect urls

### LINKED ISSUE #399: Support for a whitelist of `--redirect-domains`

boivie:
Allows having the oauth2_proxy at e.g. https://auth.example.com and using it for several of your domains, e.g. https://app1.example.com and https://app2.example.com

You will need to whitelist your domains, or set it to "*" to allow all your domains using "--redirect-domain"

Fixes #399 and #427

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,6 +21,7 @@ func main() {
 	upstreams := StringArray{}
 	skipAuthRegex := StringArray{}
 	googleGroups := StringArray{}
+	redirectDomains := StringArray{}
 
 	config := flagSet.String("config", "", "path to config file")
 	showVersion := flagSet.Bool("version", false, "print version string")
@@ -30,6 +31,7 @@ func main() {
 	flagSet.String("tls-cert", "", "path to certificate file")
 	flagSet.String("tls-key", "", "path to private key file")
 	flagSet.String("redirect-url", "", "the OAuth Redirect URL. ie: \"https://internalapp.yourcompany.com/oauth2/callback\"")
+	flagSet.Var(&redirectDomains, "redirect-domain", "Allow redirects on successful login to this domain (may be given multiple times)")
 	flagSet.Bool("set-xauthrequest", false, "set X-Auth-Request-User and X-Auth-Request-Email response headers (useful in Nginx auth_request mode)")
 	flagSet.Var(&upstreams, "upstream", "the http url(s) of the upstream endpoint or file:// paths for static files. Routing is based on the path")
 	flagSet.Bool("pass-basic-auth", true, "pass HTTP Basic Auth, X-Forwarded-User and X-Forwarded-Email information to upstream")
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -72,6 +72,7 @@ type OAuthProxy struct {
 	compiledRegex       []*regexp.Regexp
 	templates           *template.Template
 	Footer              string
+	RedirectDomains     []string
 }
 
 type UpstreamProxy struct {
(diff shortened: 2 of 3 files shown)


boivie:
If you're using Kubernetes with "external auth" in your nginx-ingress-controller, this how you configure it:

First, the oauth2-proxy. It's important to set a correct `--redirect-url`, `--cookie-domain` and `--redirect-domain`. This is an example:
```
        - --redirect-url=https://auth.internal.example.com/oauth2/callback
        - --cookie-domain=internal.example.com
        - --redirect-domain=*
```

(instead of *, you can specify one domain, or specify this multiple times with multiple domains)

In your app ingress files, specify:

```
  annotations:
    "ingress.kubernetes.io/auth-url": "http://oauth2-proxy.kube-system.svc.cluster.local:4180/oauth2/auth"
    "ingress.kubernetes.io/auth-signin": "https://auth.internal.example.com/oauth2/sign_in?rd=https://$best_http_host$request_uri"
```

This is unfortunately using the variables defined in the nginx-ingress-controller as it is right now. Better support should be added to it.

krogon-dp:
fixed #456

ploxiln:
The "any domain" `*` option should probably be removed (if not, it would require a prominent disclaimer to never use it for a real deployment). However, a "any subdomain" `*.domain.tld` option could be useful and safe.

JordanP:
@boivie in your example you set the ` --redirect-url=https://auth.internal.example.com/oauth2/callback` config option to oauth2_proxy. What If I want to secure several applications with the same oauth2_proxy instance ? I guess I don't understand what 'https://auth.internal.example.com' should point to ?

I used to have a single instance with only these options, and that used to work with nginx-ingress-controller-0.9.0-beta.11:
```
        - --provider=google
        - --email-domain=XXX
        - --upstream=file:///dev/null
        - --http-address=0.0.0.0:4180
```

janwillies:
I tested it and it works as a drop in replacement. My configuration looks exactly like @JordanP 

Image is pushed to `willies/oauth2_proxy-amd64:b7f9438_external-redirect-urls` if someone wants to try

thanks @boivie!

madmod:
Can this please be merged? Having a single oauth proxy in Kubernetes to protect various cluster sub-domains is my primary use case. I'd rather not have to make a bunch of ingresses with URL routing or something for each subdomain I want to protect.

madmod:
Actually after looking at #464 I think that better solves my issue because I would like a wildcard sub-domain rather than needing to edit the oauth proxy settings for each new ingress I want to protect.

landonwilkins:
+1, plz merge to avoid having to use a fork

ploxiln:
I suggest the alternative implementation in #464

PR #473, opened 2017-10-08
Add OpenID Connect provider name.

jmcarp:
So that the login page doesn't say "Sign in with a  provider".

cc @ericchiang

### DIFF
--- a/providers/oidc.go
+++ b/providers/oidc.go
@@ -17,6 +17,7 @@ type OIDCProvider struct {
 }
 
 func NewOIDCProvider(p *ProviderData) *OIDCProvider {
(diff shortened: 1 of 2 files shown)


ericchiang:
Isn't this going to mess up the login messages for other providers?

jmcarp:
I think this will improve messages for other providers. For example, this would change the Azure prompt from `Sign in with a Azure Account`, which is wrong (should be `an Azure Account`), to `Sign in with Azure`. Do you think this makes any messages worse?

ericchiang:
No just wanted to make sure that that was considered. "Sign in with GitHub"
and "Sign in with Google" does sound better since you mention it.

Lgtm

PR #475, opened 2017-10-09
Use SHA256 instead of SHA1

khaosjrm:
SHA1 looks like it can be exploded. SHA256 is an encryption method from SHA2, which is way more secure. This little fix will allow oauth2_proxy to use SHA2 instead and making it even more secure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ If `signature_key` is defined, proxied requests will be signed with the
 of selected request information and the request body [see `SIGNATURE_HEADERS`
 in `oauthproxy.go`](./oauthproxy.go).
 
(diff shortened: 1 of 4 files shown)


ploxiln:
The SHA methods do not do "encryption". SHA1 has not been "exploded". And for this purpose it's still perfectly fine.

https://crypto.stackexchange.com/questions/26510/why-is-hmac-sha1-still-considered-secure

khaosjrm:
well, if some day it's exploded, then... we don't need to commit new code, do we?

tcolgate:
Supporting apache apr1 , or bcrypt, may be better options.

PR #480, opened 2017-10-19
Audience flag

kfox1111:
This PR adds a flag audience-client-id to allow the passed token to be issued by a different client-id then the client-id used to generate the token.

This is used when configuring the kubernetes-dashboard. The oauth2_proxy in front of the dashboard uses the flags:
    - -client-id=kubernetes-dashboard
    - -audience-client-id=kubernetes
    - -scope=openid profile email groups audience:server:client_id:kubernetes
    - -client-secret=<kubernetes-dashboard-secret>

and in the dex config:
 - id: kubernetes-dashboard
    redirectURIs:
    - 'http://xxx.xxx.xxx.xxx:9090/oauth2/callback'
    name: 'Kubernetes Dashboard'
  - id: kubernetes
    public: true
    name: 'Kubernetes Cluster'
    trustedPeers:
      - kubernetes-dashboard

This allows the kubernetes-dashboard to have a secure redirectURL but still hand kubernetes tokens to the dashboard which in turn hands them to the kube-apiserver that expects all tokens to be issued to the client-id kubernetes.

I have validated this patch works and am prepairing an upstream chart to github.com/kubernetes/charts based on it.


jehiah:
@kfox1111 can you help me understand why you can't use the same client-id for both?

kfox1111:
When setting up dex, the cli is setup with the 'kubernetes' client id. it is setup as a public service so the cli service can use it. But the dashboard needs web based authentication in front, so making it a public service doesnt work. This allows the user to authenticate against the non public kube-dashboard client but pass a token that is intended for kubernetes to the dashboard that gives it directly to the kubeapi with the expected user. Its about the trustedPeers setting in the example.

PR #482, opened 2017-10-20
allow rewriting target prefix

brancz:
This change allows applications to that bind their HTTP endpoints normally to `/index` for example, but it can be served with a reverse proxy under `/myapp/index`, with `/myapp` being trimmed by the oauth2_proxy.

This is particularly useful when applications behind the ingress point talk to these application, and the internal applications therefore don't need to prefix their communication.

Reverse proxies often allow more generic rewriting, which we could enable here as well, but I imagined this to be one of the most common use cases, and if necessary this is open to be extended in the future.


brancz:
Rebased the PR.

PR #483, opened 2017-10-20
default to redirecting to current page

brancz:
If "-skip-provider-button" flag is used, currently a user is always
redirected to "/", whereas a user should be redirected to the page, that
was initially requested.

@jehiah


ploxiln:
consider if the sign_in page was directly visited, and see also #395

brancz:
Sorry it took so long. I addressed the sign in page redirect issue, I'd be great if this can land in 2.3. Is there an ETA?

brancz:
Friendly ping @ploxiln @jehiah :slightly_smiling_face:

ploxiln:
It looks like `OAuthStartPath` could have the same problem as `SignInPath`

PR #486, opened 2017-10-26
feat: Support proxying WebSocket connections.

chirino:


ploxiln:
there's also #201 and #145 (and others), and those two are "hundreds of lines added" rather than "thousands of lines added"

ploxiln:
oh, looks like you're copying in all dependencies. this project doesn't do that, it locks to a specific tag or commit in Godeps but does not copy them in

chirino:
Dropped the vendoring stuff, using Godeps now.

chirino:
Should I close this out? Seem the openshift fork has added websocket support and that's good enough for me.  I close this out soon if I don't here back.

chirino:
Seems there is no interest.

Issue #469, opened 2017-10-06
google group check optimization


arnottcr:
Currently the google provider uses [the list operation on the members api](https://developers.google.com/admin-sdk/directory/v1/reference/members/list) to fetch all group members in a group, then matches against the user object that has just completed the oauth flow. Google has added a [new get operation on the members api](https://developers.google.com/admin-sdk/directory/v1/reference/members/get) that takes both group id and member id, and returns a boolean. 

Moving to this new api should speed up this group membership check, that can be painful for large group sizes. I am happy to look into this change myself, but wanted to document it, in the event that I do not get a chance in the near term.

PR #500, opened 2017-11-14
Using members.get to determine group membership instead of members.list (#469)

andrewrosezen:
An initial whack at simplifying the authorization check for the Google/GSuite provider when specifying a list of allowed groups.

I've tested this informally, but so far it works as expected.


ploxiln:
Looks like a nice simplifying change.

I looked around the api a bit and I see there's a way you might be able to get all groups for a user: https://godoc.org/google.golang.org/api/admin/directory/v1#GroupsListCall.UserKey - this might involve a bit more code than what you managed here, but fewer requests to google's api.

andrewrosezen:
@ploxiln good point, the API docs suggest that this call could work fine. At a certain point, it is probably significantly less bound by http wait times. https://developers.google.com/admin-sdk/directory/v1/reference/groups/list

I'll look into this when I get a little more time.

ploxiln:
@hlhendy since it looks like you're merging stuff these days, I'd just like to point out this PR as a good little cleanup. (Although I haven't tested it myself, since I currently only use the github provider.)

andrewrosezen:
Yes, this can be merged (if it was waiting on me, sorry!)

PR #491, opened 2017-10-31
Switch from gpm -> dep for dependency management

jehiah:
This switches to use [`dep`](https://github.com/golang/dep) for dependency management.

`dep` is becoming the community standard, and solves some issues `gpm` has with windows.

This is a pretty 1:1 change, the `.travis.yml` changes to the following

```
 - wget -O dep https://github.com/golang/dep/releases/download/v0.3.2/dep-linux-amd64
 - chmod +x dep
 - ./dep ensure
```

RFR @talam @hlhendy

PR #501, opened 2017-11-16
Vendor dependencies with golang/dep

travisofthenorth:
I have a fork of this repo and `gpm` occasionally causes some grief. This PR is investigating the ability to switch over to `golang/dep` as a package manager and gauge the maintainer's interest in vendoring dependencies.


travisofthenorth:
Running into some issues with `google.golang.org/api/admin/directory/v1`, since it seems like you can only pin versions at the top-level, ie `google.golang.org/api`. In the meantime, if any maintainers can chime in on the prospect of vendoring, that would be very helpful.

jehiah:
@travisofthenorth I've started similar work in #491

My general comment is that we'd like not to commit vendored dependencies into this repo, and just the Godep metadata files, along with `./dep ensure` in the CI setup

travisofthenorth:
@jehiah awesome! I'll close this PR out then. Let me know if there's anyway I can help push your PR past the finish line.

PR #502, opened 2017-11-20
options: update options parsing for better handling of incorrect values

talam:
This is a follow-up PR to close out https://github.com/bitly/oauth2_proxy/pull/387 . Part of that PR was actually closed out by https://github.com/bitly/oauth2_proxy/pull/431 but we missed adding in some tests for it there so going to do it here.

* don't add in failed compiled regexes for skip auth regex option
* improve test coverage for skip auth regex option to handle partial
success case
* add tests for incorrect upstream options parsing errors


talam:
@jehiah this is RFR.

Issue #479, opened 2017-10-17
Feature Request: Checksum Support for Releases


BenCoffeed:
I would like to  be able to checksum verify downloaded releases to guard against MITM attacks. 
Do you currently post the checksums or have a method for verifying?

**Expected Behavior**

Check-sum provided for a release that can be calculated by the client after download and compared to posted checksum to confirm validity of downloaded code.

**Actual Behavior**

no checksum available for a release

jehiah:
👍  @BenCoffeed do you know of a Go based project that does this well?

My current workflow is to run `./dist.sh` locally on my laptop to build releases, but i know that bakes in directory information etc and isn't 100% reproducible.

BenCoffeed:
@jehiah I haven't used any personally, but [Go Releaser](https://github.com/goreleaser/goreleaser) looks neat.

talam:
Completed in https://github.com/bitly/oauth2_proxy/pull/503

PR #503, opened 2017-11-20
distribution: create sha256sum.txt file when creating version releases

talam:
This PR closes out https://github.com/bitly/oauth2_proxy/issues/479 by:
* creating a sha256sum.txt file containing sha256 checksums for each pre-built binary that we distribute with our releases. we'll have to attach this file to each release manually for now. we may or may not want to commit this file into the repo in the future, but for now this should be a good start.
* updates our README.md file with some instructions on how to verify pre-build binaries using the sha256sum.txt file we provide.


ploxiln:
hmm looks like dist.sh wasn't updated to use dep

talam:
added in an additional commit to remove the `gpm` references in `dist.sh` and replace with `dep`. RFR

PR #514, opened 2017-12-19
README: fix nginx auth_request example for requests with body

### LINKED ISSUE #442: Content-Length forwarded with auth request leading to hanging
When a POST request is made with a body, a request is made to oauth2_proxy:
```
GET /oauth2/auth HTTP/1.1
Host: [redacted]
X-Forwarded-Proto: https
X-Scheme: https
X-Auth-Request-Redirect: /api/auth/login
Content-Length: 77
Accept-Encoding: gzip
accept: application/json, text/plain, */*
cache-control: no-cache
user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3192.0 Safari/537.36
content-type: application/json;charset=UTF-8
accept-language: en-US,en;q=0.8
cookie: [redacted]
```

The GET request has a `Content-Length` header which causes oauth2_proxy to hang waiting for a body. In order to fix this we've added:
```
proxy_pass_request_body off;
proxy_set_header Content-Length "";
```
to the `location /oauth2/` block.

Does that make sense? If so, it should be added to the README to prevent others from having this same problem.

ploxiln:
Nginx never sends the body with the auth_request sub-request, but
keeps the original Content-Length header by default. Without some
config tweaks, this results in the request to /oauth2/auth hanging.

fixes #442

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,15 @@ server {
     proxy_set_header X-Scheme                $scheme;
     proxy_set_header X-Auth-Request-Redirect $request_uri;
   }
+  location = /oauth2/auth {
+    proxy_pass       http://127.0.0.1:4180;
+    proxy_set_header Host             $host;
+    proxy_set_header X-Real-IP        $remote_addr;
+    proxy_set_header X-Scheme         $scheme;
+    # nginx auth_request includes headers but not body
+    proxy_set_header Content-Length   "";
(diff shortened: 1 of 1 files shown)


talam:
LGMT

PR #517, opened 2018-01-02
If X-Original-URI header is present, use its value to check against w…

gimlet2:
…hitelist paths. That is useful for nginx auth_request mode.

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -603,6 +603,11 @@ func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) int
 	var saveSession, clearSession, revalidated bool
 	remoteAddr := getRemoteAddr(req)
 
+	originalRequestUri := req.Header.Get("X-Original-URI")
+	if originalRequestUri != "" && p.IsWhitelistedPath(originalRequestUri){
+		return http.StatusAccepted
+	}
+
 	session, sessionAge, err := p.LoadCookiedSession(req)
 	if err != nil {
 		log.Printf("%s %s", remoteAddr, err)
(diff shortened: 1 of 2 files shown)


ploxiln:
This isn't really needed for a normal nginx config ... my understanding is that this is needed for the Kubernetes nginx ingress controller?

gimlet2:
Not only k8s. But that was actually main motivation for me. If you use oauth2_proxy in pair with nginx auth_request plugin but still want to bypass auth for some endpoint you need smth like that.

gimlet2:
Right, but I also want to exclude part of api of my services to make it openly accessible.

ploxiln:
So, if using a normal nginx config, you can do that just as easily with another location block for the upstream paths you want to exclude from auth. You can do pretty much anything you want with some nginx config.

But you can't do any of that, because you're using the kubernetes ingress controller.

So the question is, wouldn't all this special adaptation just for the kubernetes nginx ingress controller, actually be better placed in the kubernetes nginx ingress controller? Why not add a no-auth whitelist feature to that? And more configuration of parameters and headers (like the `rd` parameter and the `X-Auth-Request-Redirect` header for the sign-in and start pages)? Can you provide custom nginx config snippets to the controller?

On a more practical level, you'll notice that this repo does not get much review and merging, so your chances of getting this merged are not great either way. Though I can't say for sure what will catch the maintainers' attention. I just try to clarify and filter down to the most critical PRs in the hope that it enables the most effective use of that limited resource.

gimlet2:
Beside practical motivation there is also another aspect of this pull request. The documentation of the project says `-skip-auth-regex value: bypass authentication for requests path's that match (may be given multiple times)` but that is actually very confusing because it works when you use oauth2_proxy let's say in a proxy mode, and doesn't in case of nginx auth_request approach. Yes, I can do it in k8s in another way - just make additional Ingress resource. But also would be nice to have the feature defined in docs actually working. Maybe that isn't that critical, just nice to have.

Issue #326, opened 2016-11-27
Add support for Google OAuth hd parameter


craftbyte:
When limiting login to a GApps domain, allow the user to set the `hd` parameter. This will allow Google to only display accounts on the domain.

Docs: https://developers.google.com/identity/protocols/OpenIDConnect#hd-param

craftbyte:
Example: http://cdn.craftbyte.net/Screenshot_2016-11-27-P2uIChG3.png

jehiah:
@craftbyte this sounds useful. Are you interested in contributing support for this?

craftbyte:
Sadly, I don't know GoLang, but will try :)

PR #536, opened 2018-01-29
Add option to check hosted domain for google

klautcomputing:
This implements #326 

When using the google provider and supplying a `hd` via `--hosted-domain` `oauth2_proxy` will now check for the presence of a hosted domain and return an error if not present.

Let me know if there is anything you'd like me to address code wise.


klautcomputing:
@jehiah @mbland Could I get one of you to take a look at this? That would be awesome!

PR #547, opened 2018-02-14
rd param allows other domain

icchy:
`rd` param at `/oauth2/start` currently allows only the path starts with `/`.
Otherwise the request is redirected to `/` even if rd param is required to allow other domain URI.
This commit will work on following nginx snippet:
```
location /secret {
    satisfy any;
    auth_request /oauth2/auth;
    error_page 401 = /oauth2/start?rd=$scheme://$host$uri;
}
```
which is useful for multiple path required to authenticated.


icchy:
seems to be conflicted with #464 and #544

PR #551, opened 2018-02-17
Add a Makefile and a Dockerfile which automates the build and release steps

cosmincojocar:
The following build commands are supported:

Build:
`make`

Run tests:
`make test`

Make a release:
`make release`

Make a release and build the docker image:
`make image`

Publish the docker image:
`make image-push`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM golang:1.9.4-alpine3.7
+
+ENV BIN=oauth2_proxy
+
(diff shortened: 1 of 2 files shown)


pataquets:
Strongly against Docker image publishing via ```docker push``` vs. using [Automated Builds from source on Docker Hub](https://docs.docker.com/docker-hub/builds/) as an official Docker release method.
Automated Builds create image tags from git tags & branches and provide code traceability, opposed to ```docker push```'ing.
See #372 and #460.

PR #556, opened 2018-03-02
Respect the X-Forwarded-Proto header

tomasbasham:
The `X-Forwarded-Proto` header is the standard way to identify over what protocol a request has been made by a client to a proxy or load balancer. Some load balancers (including Google Cloud Load Balancer - one which I use) set this header before forwarding the request.

I have setup `oauth2_proxy` to sit behind the GCLB (that performs TLS termination) however currently when I hit the load balancer over HTTP it does not redirect to HTTPS before initiating the OAuth dance causing the authentication to fail due to the mismatch of protocol. To prevent this I have had to sit nginx in front of `oauth2_proxy` just to do the HTTPS redirection. The current infrastructure look like the following:

    GCLB -> nginx -> oauth2_proxy -> app

This PR introduces the ability for the `oauth2_proxy` to handle the redirection by respecting the `X-Forwarded-Proto` header removing the need for an extra proxy (nginx or similar).


bravoecho:
Hi, it would be lovely if anyone could help with the `dep ensure` failing on Travis. 

We would then greatly appreciate if this PR was considered for review and merging because at the moment we are using a fork, and if the functionality doesn't get into upstream we will be forced to revert to an needlessly complicated nginx setup sidecar in our deployment. 

Thank you!! 🙂

ploxiln:
re: `dep ensure`: see #574 and #595 - the inotify situation is confusing. your local dep may have cached the inotify import _metadata_ which enables it to continue working. On completely fresh systems it doesn't _quite_ work.

tomasbasham:
Thanks @ploxiln. I guess there is noting I can do until either of these gets merged save me having to duplicate the effort.

PR #557, opened 2018-03-05
Add support for AWS request signing

chrissnell:
This adds support for signing upstream requests with an AWS-style `Authorization` header.  This is used for AWS services like AWS Elasticsearch and AWS Kibana, which can be configured to require the client to present a valid authorization token.

To use this feature, you need to have your `~/.aws` directory configured (using the AWS CLI client) or set the `AWS_ACCESS_KEY` and `AWS_SECRET_ACCESS_KEY` environmental variables.  You will also need to pass two flags, `sign-aws-request-region` and `sign-aws-request-service`, which specify the AWS service and region you are attempting to access on the upstream request.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -4,8 +4,8 @@
 [[projects]]
   name = "cloud.google.com/go"
   packages = ["compute/metadata"]
-  revision = "767c40d6a2e058483c25fa193e963a22da17236d"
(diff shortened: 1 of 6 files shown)


TGNA:
Hi @chrissnell, I'm interested in your work. Why did you close the MR? Isn't it possible to integrate it?

Issue #560, opened 2018-03-09
error redeeming code: illegal base64 data


danopia:
I use oauth2_proxy with a custom provider, using login_url, redeem_url, and profile_url. Recently, new proxies stopped working, and simply 500'd when I tried logging in.

I traced this back to #421. My provider puts padding on the JWTs and padding is no longer allowed. I've rolled back to a working version, but would appreciate guidance on how to fix this for the future.

---

```
oauth2_proxy
  -upstream
  http://localhost:3000
  -config
  /conf/oauth2_proxy.cfg
  -skip-provider-button
```

oauth2_proxy.cfg:

```
login_url = "https://custom-server/oauth2/auth"
profile_url = "https://custom-server/api/v1/me"
redeem_url = "https://custom-server/oauth2/access_token"
email_domains = ["*"]
http_address = "http://0.0.0.0:4180"
pass_basic_auth = false
```

danopia:
Since my patch for this issue was merged back in March, I'm closing my own issue here.

The actual discussion (however minimal) happened in #561

PR #561, opened 2018-03-09
Strip JWT base64 padding before parsing. #560

danopia:
This provider used to allow padding, until it was switched to the builtin RawURLEncoding. This PR fixes the regression by accepting padding again via throwing the padding away.

Google doesn't include padding, but other OAuth2 servers may, and this provider is the default when using a custom provider.

### DIFF
--- a/providers/google.go
+++ b/providers/google.go
@@ -67,7 +67,8 @@ func emailFromIdToken(idToken string) (string, error) {
 	// id_token is a base64 encode ID token payload
 	// https://developers.google.com/accounts/docs/OAuth2Login#obtainuserinfo
 	jwt := strings.Split(idToken, ".")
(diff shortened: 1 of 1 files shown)


ploxiln:
I think #421 was a mistake - we did not intent to stop supporting padding. It should probably just be reverted. cc @talam

talam:
Sorry, that's a mistake on my part. cc @jehiah who can take a look as I am not a maintainer anymore.

If we don't end up reverting the commit, it would be good to include a test case for this as well in `google_test.go`

ploxiln:
... now that I look closer, this is sort of equivalent reverting #421, but a bit simpler IMHO, so I change my mind and think we should merge this :)

Issue #568, opened 2018-03-20
gopkg.in/fsnotify.v1 import path is broken


btuan:
Causes issues with `go get -u github.com/bitly/oauth2_proxy`:
```
$ go get github.com/bitly/oauth2_proxy
package gopkg.in/fsnotify.v1: unrecognized import path "gopkg.in/fsnotify.v1" (parse https://gopkg.in/fsnotify.v1?go-get=1: no go-import meta tags ())
```

`fsnotify` dev team posted:
>fsnotify 
> @fsnotify
Follow Follow @fsnotify
More
If you are still using http://gopkg.in  for fsnotify, please update your import path.
http://github.com/fsnotify/fsnotify … #golang

https://twitter.com/fsnotify/status/976222199752224768

btuan:
Closing issue as `gopkg.in` redirect is patched:
https://github.com/go-fsnotify/fsnotify/issues/1#issuecomment-374967614

blaskovicz:
@btuan wouldn't the import still need updating?

```
$ go get ./...
package gopkg.in/fsnotify.v1: unrecognized import path "gopkg.in/fsnotify.v1" (parse https://gopkg.in/fsnotify.v1?go-get=1: no go-import meta tags (meta tag gopkg.in/fsnotify/fsnotify.v1 did not match import path gopkg.in/fsnotify.v1))

$ go build
watcher.go:11:2: cannot find package "gopkg.in/fsnotify.v1" in any of:
        /usr/local/go/src/gopkg.in/fsnotify.v1 (from $GOROOT)
        /home/blaskovicz/src/gopkg.in/fsnotify.v1 (from $GOPATH)
```

btuan:
@blaskovicz yes, you're right!

PR #569, opened 2018-03-20
Gopkg: Fix fsnotify import path

btuan:
RE: Issue #568 

fsnotify dev team posted:
twitter.com/fsnotify/status/976222199752224768

Example error manifestation:
```
$ wget https://gopkg.in/fsnotify.v1
--2018-03-20 16:15:33--  https://gopkg.in/fsnotify.v1
Resolving gopkg.in (gopkg.in)... 45.33.37.13
Connecting to gopkg.in (gopkg.in)|45.33.37.13|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2018-03-20 16:15:35 ERROR 404: Not Found.
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -98,7 +98,7 @@
   version = "v1.0.0"
 
 [[projects]]
-  name = "gopkg.in/fsnotify.v1"
+  name = "gopkg.in/fsnotify/fsnotify.v1"
   packages = ["."]
   revision = "836bfd95fecc0f1511dd66bdbf2b5b61ab8b00b6"
   version = "v1.2.11"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -36,5 +36,5 @@
   name = "google.golang.org/api"
 
 [[constraint]]
-  name = "gopkg.in/fsnotify.v1"
+  name = "gopkg.in/fsnotify/fsnotify.v1"
   version = "~1.2.0"
--- a/watcher.go
+++ b/watcher.go
@@ -8,7 +8,7 @@ import (
 	"path/filepath"
 	"time"
 
-	"gopkg.in/fsnotify.v1"
+	"gopkg.in/fsnotify/fsnotify.v1"
 )
 
 func WaitForReplacement(filename string, op fsnotify.Op,


Ginja:
👍 

Perhaps the import path should be the github.com URL of the project (github.com/fsnotify/fsnotify)? It also seems that the maintainer of fsnotify prefers this.

https://github.com/go-fsnotify/fsnotify/issues/1#issuecomment-375413181
https://github.com/fsnotify/fsnotify/issues/108

ldionmarcil:
Hi. I was in the process of reporting this to affected organizations. Yes, you can mitigate this by pointing directly to github.com/fsnotify/fsnotify. You can still use gopkg.in if you need a specific version of this project in order to fetch the repository but be sure to always use gopkg.in/{org_name}/{repo_name}.

niemeyer:
@ldionmarcil Please stop spreading this misinformation. That's not good for you or for anybody else, and makes me believe less in the theory that you intended well by breaking everybody.

There's zero benefit in using foo/bar vs. simply foo for repositories that are named go-foo/foo. The problem that affected fsnotify occurred because fsnotify _renamed their repository_ from "go-fsnotify" to "fsnotify". If they were using "gopkg.in/go-fsnotify/fsnotify" as the import path or even "github.com/go-fsnotify/fsnotify" and renamed the repository, the exact same problem would happen.

ldionmarcil:
I think there is value in mentioning that `gopkg.in` will append `go-` to the repository name if it is not specified. This is the sort of behaviour that can lead developers into not realizing (or forgetting) that using the `gopkg.in/{repo_name}` import value is actually pointing to `go-{repo_name}`. 

Had the developers been using `gopkg.in/{org_name}/{repo_name}` initially, they would have noticed the break instantly, instead of leaving an insecure open redirect dormant for literally two years. You are free to think I am saying nonsense, but I will continue to state that the default behaviour of `gopkg.in` is weird and it will definitely lead to this type of scenario again in the future.

niemeyer:
@ldionmarcil The entirety of your message is incorrect, @ldionmarcil. You don't understand even at a trivial level how gopkg.in works (hint: there's documentation about it in the front page), and also don't understand why that github.com URL has exactly the same problem. Please don't create more damage than you already did and do a bit of home work, Louis.

blaskovicz:
For anyone that `go get`s this and wants to fix it locally until this is merged:

1) `$ sed -i 's@gopkg.in/fsnotify.v1@gopkg.in/fsnotify/fsnotify.v1@' watcher.go Gopkg.toml Gopkg.lock`
2) `$ go get ./...`
3) `$ go build`
4) `$ ./oauth2_proxy`

niemeyer:
There's no benefit @blaskovicz, as gopkg.in itself got a workaround for the rename. The code will be the same with or without that local change.

btuan:
Closing PR due to fixed redirect:
https://github.com/go-fsnotify/fsnotify/issues/1#issuecomment-374967614

btuan:
pulled the trigger too quickly and deleted this branch, but the patch given in this PR should still suffice to fix:
```
btuan$ go get -u gopkg.in/fsnotify.v1
package gopkg.in/fsnotify.v1: unrecognized import path "gopkg.in/fsnotify.v1" (parse https://gopkg.in/fsnotify.v1?go-get=1: no go-import meta tags (meta tag gopkg.in/fsnotify/fsnotify.v1 did not match import path gopkg.in/fsnotify.v1))
```

niemeyer:
@btuan My apologies.. I got the redirect logic wrong the first time. This should be fixed now.

Issue #562, opened 2018-03-13
RequestURI prepend


rmazlumyan:
Would be useful to have a simple rewrite like option to prepend a path to the RequestURI before proxying upstream.

For instance, an incoming request for `/index.html` can be rewritten to `/my-prepend/index.ytml` when sent upstream.

I can submit a PR if there's interest.

cliveza:
In my case I want put the proxy in front of Kibana, that's host on AWS as part of there Elastic Search Service, so that all request to `https://kibana.mydomain.com` target the upstream as `https://vpc-es-11112222333444111-east-1.es.amazonaws.com/_plugin/kibana/`

Its the `/_plugin/kibana/` that I want to automatically add if it's **not there already**, seems like this change will add that.

PR #579, opened 2018-04-10
Option to prepend a path to the RequestURI before proxying upstream

rmazlumyan:
Adds -rewrite-prepend option to prepend string to request URI before proxying to upstream

#562


cliveza:
👍 I need this option too please.

PR #585, opened 2018-05-10
Fixed nginx 401 redirect

SuperSandro2000:
The 401 redirect in the nginx example is broken in the latest version of nginx.
Ref: https://www.digitalocean.com/community/tutorials/how-to-configure-nginx-to-use-custom-error-pages-on-ubuntu-14-04#configuring-nginx-to-use-your-error-pages

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,7 +399,7 @@ server {
 
   location / {
     auth_request /oauth2/auth;
-    error_page 401 = /oauth2/sign_in;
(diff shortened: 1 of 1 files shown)


elupus:
Why is it broken? The syntax means different things. Original means, let oauth proxy decide on the http status code (200), your mean keep 401, but content from oauth proxy.

PR #587, opened 2018-05-10
Sets the redirect URL default to the original request's path instead …

### LINKED ISSUE #586: Redirect URL missing when the skip-provider-button option is used
Hi folks :wave: 

At my organization we are using the oauth2_proxy to provide authentication via GitHub for internal sites.
One issue we ran into is that, when skipping the sign-in page to directly go to `/oauth/start`, the redirect URL that gets sent to the auth provider (in our case, GitHub) via the `state` query string parameter, is always set to "/".

It is being set by the `GetLoginURL` function:
https://github.com/bitly/oauth2_proxy/blob/master/oauthproxy.go#L521
The state is obtained by calling the GetRedirect function on the http request:
https://github.com/bitly/oauth2_proxy/blob/master/oauthproxy.go#L428

The issue is that when going directly to the provider's authentication page, the `rd` form parameter that the function tries to parse never gets set in the first place: https://github.com/bitly/oauth2_proxy/blob/master/templates.go#L114, so the redirect path defaults the root path:
https://github.com/bitly/oauth2_proxy/blob/master/oauthproxy.go#L430

Which ends up redirecting users to the root path after they have been authorized to access the page they requested (i.e. `/` instead of `/internal-page`).

Would it be reasonable to, instead of defaulting to "/", try and set the redirect variable to the path the user landed on when beginning the auth cycle, so that they can land back on the page they requested in the first place?

This is what ended up working for us:
```
diff --git a/oauthproxy.go b/oauthproxy.go
index 21e5dfc..2981a9f 100644
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -427,7 +427,7 @@ func (p *OAuthProxy) GetRedirect(req *http.Request) (redirect string, err error)
 
        redirect = req.Form.Get("rd")
        if redirect == "" || !strings.HasPrefix(redirect, "/") || strings.HasPrefix(redirect, "//") {
-               redirect = "/"
+               redirect = req.URL.Path
        }
 
        return

```
I am opening a PR in case the approach sounds reasonable to you.

Thank you:)

owayss:
Fixes #586

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -427,7 +427,7 @@ func (p *OAuthProxy) GetRedirect(req *http.Request) (redirect string, err error)
 
 	redirect = req.Form.Get("rd")
 	if redirect == "" || !strings.HasPrefix(redirect, "/") || strings.HasPrefix(redirect, "//") {
-		redirect = "/"
+		redirect = req.URL.Path
 	}
 
 	return


oauthproxy.go
```diff
@@ -427,7 +427,7 @@ func (p *OAuthProxy) GetRedirect(req *http.Request) (redirect string, err error)
 
 	redirect = req.Form.Get("rd")
 	if redirect == "" || !strings.HasPrefix(redirect, "/") || strings.HasPrefix(redirect, "//") {
-		redirect = "/"
+		redirect = req.URL.Path
```
ploxiln: this does not work out because if you visited `/oauth2/sign_in` or `/oauth2/start` explicitly, then you end up in a loop

telnet2: What about using like this?

```
if req.Header.Get("X-Auth-Request-Redirect") != "" {
    redirect = req.Header.Get("X-Auth-Request-Redirect")
}
```

PR #574, opened 2018-03-26
Fix fsnotify package name

jj3l:
Fix for https://github.com/bitly/oauth2_proxy/issues/573

Issue #573, opened 2018-03-26
go get github.com/bitly/oauth2_proxy  Does not work


grebenaha:
go get github.com/bitly/oauth2_proxy 
**package gopkg.in/fsnotify.v1: unrecognized import path "gopkg.in/fsnotify.v1" (parse https://gopkg.in/fsnotify.v1?go-get=1: no go-import meta tags (meta tag gopkg.in/fsnotify/fsnotify.v1 did not match import path gopkg.in/fsnotify.v1))**

ploxiln:
related: #568 / #569

PR #595, opened 2018-05-16
update fsnotify package name, as recommended by fsnotify doc

shamil:
Fix for #573 to make `dep ensure` work 

Similar to #574, but uses package name (`github.com/fsnotify/fsnotify`) as recommended [here](https://godoc.org/github.com/fsnotify/fsnotify)
Their example in https://github.com/fsnotify/fsnotify/blob/master/example_test.go also use `github.com/fsnotify/fsnotify` as package name

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -32,6 +32,12 @@
   version = "v1.1.0"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


shamil:
closing due to #628

PR #607, opened 2018-05-29
github: check if email is verified

caarlos0:
I believe its more important that the email is
verified than that it is the primary email.

I can have several emails with different domains,
and I could want to login in different oauth2_proxy
instances using different domains as filters.


caarlos0:
(don't know why `TestValidatorEmpty` is failing on travis)

PR #613, opened 2018-06-03
Fix GitHub teams for large orgs

skwashd:
Some large organisations have more than 100 teams. The current
implementation only checks the first 100 teams returned by the
API. This checks all teams the user is a member of.

Added support for GitHub's new nested team feature which means
inheritence works as designed. Again useful for large orgs.

Updated the limit from 200 to the [documented limit of
100](https://developer.github.com/v3/#pagination).

### DIFF
--- a/providers/github.go
+++ b/providers/github.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


skwashd:
Travis failing due to an error with dep. https://travis-ci.org/bitly/oauth2_proxy/jobs/387385489#L598 The last 3 master builds have failed due to issues with fsnotify.v1.

Issue #384, opened 2017-04-28
UI Redressing (Clickjacking)


permyakovsv:
Add X-Frame-Options: SAMEORIGIN to the existing response headers or make it optional.

jehiah:
@permyakovsv this would be a great addition. Would you like to contribute a PR?

This should probably be standard for the oauth2 sign in page responses (at least i can't think of a good reason why it shouldn't be), and should be optional for all responses.

reedloden:
Might be good to integrate something like https://github.com/unrolled/secure for more than just `X-Frame-Options` support.

PR #618, opened 2018-06-18
Add secure middleware to enable more header options

timothy-spencer:
I needed to be have XSS and nosniff headers set to help compliance scans not complain, so I followed through with the suggestion in https://github.com/bitly/oauth2_proxy/issues/384.  

Please let me know if you have any suggestions or things that I could do to help this get merged in.  We would love to be able to just use upstream for this.  

Thanks!

### DIFF
--- a/oauthproxy_test.go
+++ b/oauthproxy_test.go
@@ -863,5 +863,6 @@ func TestHttpBrowserXssFilterFalseBydefault(t *testing.T) {
 	rw := httptest.NewRecorder()
 	req, _ := http.NewRequest("GET", "/", nil)
 	proxy.ServeHTTP(rw, req)
+	assert.Equal(t, false, opts.HttpBrowserXssFilter)
 	assert.Equal(t, "", rw.HeaderMap.Get("X-XSS-Protection"))
 }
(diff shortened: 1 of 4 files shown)


oauthproxy_test.go
```diff
@@ -836,3 +837,73 @@ func TestRequestSignaturePostRequest(t *testing.T) {
 	assert.Equal(t, 200, st.rw.Code)
 	assert.Equal(t, st.rw.Body.String(), "signatures match")
 }
+
+func TestHttpBrowserXssFilterTrue(t *testing.T) {
```
jmcarp: It looks like this tests a third-party package but doesn't really test the behavior of this package. What kinds of regressions would this test catch?

timothy-spencer: Well, other tests further up in the file test the proxy.  This tests that the middleware is functioning on top of the proxy.  I definitely don't have 100% coverage here.  I only am testing one of the secure middleware options, for example.  More tests certainly could be written!

These were sufficient for me to verify that the proxy with middleware worked for our purposes, though.

jmcarp: I'm not sure this tests *any* of the code you added. As far as I can tell, the test imports a third-party package, instantiates its middleware, and verifies that the middleware behaves as expected. It doesn't really exercise the code you added, or any of the code in this package.

timothy-spencer: Yes, that's how main() does it too.  Remember how you told me to wrap the middleware outside of NewOAuthProxy?  This is the result.  As far as I can see, this is the only way I can actually do this sort of thing.  As a golang neophyte, I welcome being proven wrong.

PR #619, opened 2018-06-19
Use insecure transport if SSLInsecureSkipVerify is set to true

### LINKED ISSUE #403: Can't skip validation of self signed certificate
In my development environment my upstream server has a self signed certificate. I have tried using the --ssl-insecure-skip-verify=true option, but I still receive the following error:

2017/06/12 16:46:22 reverseproxy.go:316: http: proxy error: x509: certificate signed by unknown authority

hrenod:
Fixes #403

### DIFF
--- a/options.go
+++ b/options.go
@@ -132,6 +132,7 @@ func (o *Options) Validate() error {
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 		}
 		http.DefaultClient = &http.Client{Transport: insecureTransport}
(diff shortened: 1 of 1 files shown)


funkypenguin:
Could we have another attempt at this CI test? The Travis logs don't seem to have anything to do with the diff in the commit, could this have been a transient fault (I'd also like #403 fixed!) :)

ploxiln:
failed due to needing fsnotify import fix, see #574

PR #627, opened 2018-07-05
optionally expose the access token in a response header

bodewig:
This is similar to what #621 does to the id_token but exposes the access_token.

Use case is an nginx with `auth-request` that wants to make the access token available to the resource server "behind" the nginx.


ploxiln:
and very similar to #424 ;)

bodewig:
argh, sorry about that. I did look for open issues and PRs before I started working on this but missed #424.

Maybe the testcase of this PR could be adapted and added to #424 ?

I'll close this one as there is little reason to add more noise to an already long list of PRs.

PR #636, opened 2018-07-25
Add freebsd as a build target

jyundt:


jyundt:
Would it be possible to add FreeBSD as a build target? I did a (very) quick test and built oauth2_proxy-2.2.1-alpha using golang 1.10.3.

It seems to run fine on FreeBSD but I need to test a bit more.

fnkr:
We are using this on FreeBSD and it works just fine. I would like to have this merged too.

PR #649, opened 2018-09-07
Add flag to skip upstream validation and proxying.

vharish836:
I wanted to use oauth2_proxy along with auth_request module of NGINX for handling multiple sub-domains. In this case upstream is not a mandatory config since nginx is doing the proxying an our job is only to authenticate.

I did not find any way of avoiding upstream with current implementation and hence made this change.


ploxiln:
you can simply not specify any upstreams, since https://github.com/bitly/oauth2_proxy/pull/426

PR #661, opened 2018-10-18
RD-4503: Handle missing username field in Gitlab

gubbin:
For normal users, Gitlab has an email and username field in the session info.  Admin users only have email.


gubbin:
Meant to PR this into the CivicActions fork.

PR #676, opened 2018-12-01
Redirect to uri even if --skip-provider-button

leon-barrett:
The normal behavior is that if a user is going to page
https://hostname/x but is not authenticated, they are sent to login and
then redirected back to https://hostname/x.

I discovered that if the flag --skip-provider-button is given, then
after login the user ends up at https://hostname/.

I believe this to be a bug. That behavior seems to happen because the
current URI is only extracted when redirecting to the sign-in page form,
not when the OAuth flow is started, so if that form is skipped, the
current URL is lost. Therefore, I made a change to extract the current
page to send into the OAuth flow.

This change fixes that bug and adds a test.

### DIFF
--- a/oauthproxy.go
+++ b/oauthproxy.go
@@ -469,7 +469,7 @@ func (p *OAuthProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	case path == p.SignOutPath:
 		p.SignOut(rw, req)
 	case path == p.OAuthStartPath:
-		p.OAuthStart(rw, req)
+		p.OAuthStart(rw, req, "")
 	case path == p.OAuthCallbackPath:
 		p.OAuthCallback(rw, req)
 	case path == p.AuthOnlyPath:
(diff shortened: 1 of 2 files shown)


leon-barrett:
While the tests failed, they seemed to fail on dependencies (none of which I touched).

https://travis-ci.org/bitly/oauth2_proxy/jobs/462032033#L588
> ensure Solve(): No versions of gopkg.in/fsnotify.v1 met constraints:

https://travis-ci.org/bitly/oauth2_proxy/jobs/462032032#L570
> ../../../golang.org/x/oauth2/google/default.go:38: syntax error: unexpected = in type declaration

I can investigate what's needed to update those dependencies, but I'd naively expect that you'd have more luck :)

leon-barrett:
Ah, it looks like https://github.com/bitly/oauth2_proxy/pull/610 is at least a partial fix.

ploxiln:
this repo is unmaintained, see https://github.com/bitly/oauth2_proxy/issues/628#issuecomment-417121636

